PR #18, opened 2026-05-03
feat: full pentest tool suite and modular recon pipeline (#13)

### LINKED ISSUE #13: feat: ensure all agents have adequate tooling
I am concerned that some agents don't have tools at all. Should they?

Some ideas:
- disclosure coordinator might want to tweet
- technical author might want to produce more than .md
  - is that the whole submission API or do we want to submit/does H1 accept pics/logs too?
- vulnerability researcher might want some research tools
- not one RSS?
- surely there are more tools for the osint analyst, penetration tester, and vulnerability researcher?

coolhandle01:
Closes #13.

## What this does

Replaces the original stub tool set with a complete, agent-driven security pipeline. Every agent now has real, scoped tools - the Penetration Tester in particular went from 2 tools (nuclei, sqlmap) to 28 targeted checks it selects based on recon evidence rather than running everything blindly.

## Recon (`tools/recon/`)

Modular rewrite replacing the old monolithic `recon_tools.py`:

- subfinder, httpx, nmap, waybackurls, cert transparency (existing, now modular)
- `dirfuzz` - ffuf wrapper with bundled wordlist
- `tls` - testssl.sh wrapper for TLS config assessment; dnspython-backed SPF/DMARC checks
- `traceroute` - tracepath/traceroute wrapper storing hop IPs in `ReconResult.network_hops` for CDN/WAF bypass reasoning

TLS and DNS findings are collected during recon as `ReconResult.passive_findings` so all downstream agents have this context without the PT needing to re-run them.

## Pentest (`tools/pentest/`)

Modular rewrite replacing the old `vuln_tools.py`. The monolithic `run_pentest()` orchestrator is gone - the PT agent now makes its own decisions using individual targeted tools:

- nuclei with technology-tag filtering (`-tags wordpress,cve` etc.)
- sqlmap, CORS, SSRF, reflected XSS, SRI, CRLF/header injection, host header attacks
- JS source map discovery + gitleaks-backed secret scanning
- Error and stack trace disclosure probing (15 patterns: Python, PHP, Java, Node, SQL dialects)

## Cloud (`tools/cloud/`)

New package with three sub-areas:

- `aws.py` - S3 public bucket listing and accessibility
- `azure.py` - Azure Blob container listing and SAS token detection
- `services.py` - sensitive files, admin panel paths, branded panels (cPanel/WHM, Plesk, DirectAdmin, Webmin, Grafana, Kibana, Portainer, Consul/Vault)
- `databases/` - per-engine unauthenticated checks: Elasticsearch (HTTP), CouchDB (HTTP), Redis (PING), MongoDB (wire protocol isMaster), PostgreSQL (startup/trust-auth detection), MySQL (greeting check)

## PT agent tool design

Tools that benefit from selectivity take `endpoints_json` (forcing the agent to extract specific endpoints) rather than the full `ReconResult`. Tools that need port or subdomain context take `recon_result_json`. All 28 tools have docstrings that tell the agent exactly when to use them.

## Programme Manager enrichment

- `Programme` model: `offers_bounties`, `accepts_new_reports`, `response_efficiency_pct`, `avg_time_to_bounty_days`, `total_bounties_paid_usd`, `policy_text`
- Hard pre-filters: VDPs and closed programmes rejected before scoring
- Raw `policy_text` passed to the LLM for rules-of-engagement evaluation
- 40/20/20/20 scoring model (payout ceiling / attack surface / programme health / response speed)

## Tests

347 unit tests, 85% coverage. Every new tool function has a corresponding test; all external binaries and network calls are mocked.


coolhandle01:
Moar tools!

There's a bunch of one-liners here that might make nice crew.ai Tools:
https://github.com/dwisiswant0/awesome-oneliner-bugbounty/blob/master/README.md

I think more header attacks could be tested:
https://chennyren.notion.site/HTTP-Host-header-attacks-174eab8670e347348079164548008988

I'm not seeing anything about checking for .js source maps, reconstructing sources and looking for tokens/URLs

Nothing for Websocket stuff, the tools list here looked interesting:
https://github.com/PalindromeLabs/awesome-websocket-security/blob/main/README.md

Nothing for waf bypass, nothing to safely test for RCE in PHP, React, or expressjs.. nothing for web assembly, nothing for writable s3 buckets or writable azure blob storage.. we should probably have toolkits to audit any cloud service resources we encounter.. how about them SAS tokens? What about unauthenticated databases? Exposed admin panels?

It seems there are one or two folks documenting recon workflows on github to aid beginners. Repos like this should be evaluated to figure out what a good, reliable, exhaustive recon workflow looks like.

https://github.com/Mehdi0x90/Web_Hacking/blob/refs%2Fheads%2Fmain/Recon.md
https://github.com/AdmiralGaust/bountyReconV2
https://github.com/gokulapap/Reconator
https://github.com/bl4de/security-tools/tree/master
https://github.com/RevoltSecurities/Subdominator
https://github.com/trustedsec/ptf
https://github.com/darkoperator/dnsrecon
https://github.com/nabla-c0d3/sslyze


What about skills?
https://github.com/shuvonsec/public-skills-builder
https://github.com/Orizon-eu/claude-code-pentest

-----

I wonder what benefits and drawbacks there are to building more monolithic crew.ai tools that encapsulate the most high performance bug bounty workflows, versus providing a fuckton of tiny crew.ai tools that are parameterised wrappers of one-liners and letting agents build their own workflows OTF?

I think I would err towards being opinionated with crew.ai tool design because it should save tokens both per run and over time (agents won't need to relearn how tools plug into each other, and agent .md files will not have to describe all of them), plus we can define a sort of "best practice" (clearly I don't know what best practice is! ;)), that even if it isn't perfect all the time, is Good Enough

Maybe I haven't considered all aspects though. For example, maybe the pipeline of Tasks doesn't reflect "doing a bug bounty" very well. I am open to evaluating the crew.ai Task pipeline Tasks and execution order. -- maybe recon is better as a Tool rather than a Task with a dedicated agent to execute it.

-----

I think it might be time to make tools more modular by task/tool, like tools/recon/waybackurls.py, tools/pentest/webapp_headers.py, tools/recon/webapp_body.py.

-----

I think it's important that the cybersquad's success rate at finding "the obvious stuff" like writeable S3 buckets is very high. They squad should never miss that sort of low-hanging fruit, ideally logging it before anyone else! This is why I think it's important that the agents have tooling that enables finding that low hanging fruit efficiently and reliably.

tools/recon/probe.py
```diff
@@ -0,0 +1,46 @@
...
+from config import config
+from models import Endpoint
+from tools._helpers import _require_binary, _run
+
+logger = logging.getLogger(__name__)
+
+
+def probe_endpoints(hosts: list[str]) -> list[Endpoint]:
+    """Use httpx CLI to probe a list of hostnames/URLs for live services."""
+    httpx_bin = _require_binary("httpx")
```
coolhandle01, line 21: I think we use requests library in some of the other tools. We should probably choose the requests library or httpx but not both. I am not familiar with httpx so I would naively prefer just using the requests library unless httpx provides something that requests doesn't

tests/test_recon_tools.py
```diff
@@ -334,3 +336,97 @@ def test_empty_output_returns_empty_list(self):
...
+
+        urls = [ep.url for ep in result]
+        assert "https://example.com/admin" not in urls
+        assert "https://example.com/new-path" in urls
+
+    def test_empty_endpoints_returns_empty(self):
+        assert discover_paths([]) == []
+
+    def test_skips_endpoints_with_500_status(self):
+        endpoints = [Endpoint(url="https://example.com/", status_code=500)]
```
coolhandle01, line 393: It occurs to me that when we're doing all this recon, in cases where we do get an error, we probably want to check that a debug stack trace hasn't leaked through, maybe the stack trace will lead to something exploitable.  -- https://trufflesecurity.com/blog/claude-tried-to-hack-30-companies-nobody-asked-it-to

squad/penetration_tester/__init__.py
```diff
@@ -6,25 +6,96 @@
...
+    """
+    Discover .js.map source map files linked from live HTML pages, parse them, and
+    scan reconstructed source code for leaked API keys, tokens, passwords, and
+    internal file paths.
+    """
+    recon = ReconResult.model_validate_json(recon_result_json)
+    return [f.model_dump() for f in check_js_source_maps(recon.endpoints)]
+
+
+@tool("Cloud Misconfiguration Check")
```
coolhandle01: I think this should be three separate tools.. I need to look into how we're using these checks: i don't want to blindly scan azure IP spaces but rather make sure that we are finding e.g. S3 or blob URLs with SAS tokens in e.g. reconstructed js, or if we discover that some html is served from some blob storage account (think azure blob storage hosted website with an azure tls cert and domain), or some iframe, or some other thing beyond my imagination..

squad/penetration_tester/__init__.py
```diff
@@ -6,25 +6,96 @@
...
+    return [f.model_dump() for f in findings]
+
+
 MEMBER = SquadMember(
     slug="penetration_tester",
     dir=Path(__file__).parent,
-    tools=[pentest_tool],
+    tools=[
+        pentest_tool,
+        ssrf_probe_tool,
```
coolhandle01, line 456: Moar abbreviations! XD

I think we should probably check for xss as we do ssrf with a little python if we can. Nuclei probably does xss checks? But still it should be easy enough to implement and it means the pentester can use it selectively.

tests/test_sourcemaps.py
```diff
@@ -0,0 +1,193 @@
...
+                resp.text = json.dumps(map_data)
+                resp.json.return_value = map_data
+            elif url.endswith(".js"):
+                resp.text = ""
+            else:
+                resp.text = _html_with_script()
+            return resp
+
+        with patch("requests.get", side_effect=fake_get):
+            results = check_js_source_maps([ep])
```
coolhandle01, line 213: I don't think I want to solve this problem.. Gitleaks is the answer here methinks

squad/penetration_tester/__init__.py
coolhandle01: Like. There's a reason I have to put a checksum on my CSS links in <head> but I can't remember why xD

tools/cloud/azure.py
```diff
@@ -0,0 +1,91 @@
+"""Azure Blob Storage misconfiguration checks."""
```
coolhandle01, line 1: I'd like to rename this to azure.py so that it's a natural home to expand for other azure services later like databases, serverless weirdness and so much more eventually I guess

tools/cloud/azure.py
```diff
@@ -0,0 +1,91 @@
...
+                        "SAS token query parameters (sv/se/sig/sr/sp) detected in URL. "
+                        "Tokens embedded in URLs are logged by proxies and browsers."
+                    ),
+                    tool="azure_storage_check",
+                    severity_hint=Severity.HIGH,
+                )
+            )
+
+    # Check for publicly listable containers
+    for account in _account_candidates(recon):
```
coolhandle01, line 66: This should really be about processing results/links from within scope found in Ajax type stuff (reconstructed js, resourse links for or CSS, or some header self-snitches that the site is hosted on blob storage), but it looks like it's basically trying to guess scope. Bit scary.

tools/pentest/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    "is_in_scope",
+    "lookup_cve",
+    "run_nuclei",
+    "run_pentest",
+    "run_sqlmap",
+    "triage_findings",
+]
+
+
+def run_pentest(recon: ReconResult) -> list[RawFinding]:
```
coolhandle01: You're right. Whilst we can use heavyweight automated tools to do the build of the work, we need to allow the penetration tester and vulnerability hunter and opportunity to validate the those tools findings and to also maybe get creative (within reason). They need to be able to figure out if findings can lead to more findings..

tools/pentest/sourcemaps.py
```diff
@@ -0,0 +1,184 @@
...
+from config import config
+from models import Endpoint, RawFinding, Severity
+
+logger = logging.getLogger(__name__)
+
+_JS_SRC_RE = re.compile(r'<script[^>]+src=["\']([^"\']+\.js(?:\?[^"\']*)?)["\']', re.I)
+_SOURCEMAP_COMMENT_RE = re.compile(r"//# sourceMappingURL=(.+)$", re.M)
+
+# Pattern tuples: (compiled_regex, label)
+_SECRET_PATTERNS: list[tuple[re.Pattern[str], str]] = [
```
coolhandle01: Nahhh definitely we should let gitleaks or similar handle all this stuff. It's hard.

tools/pentest/ssrf.py
```diff
@@ -0,0 +1,61 @@
...
+
+import requests
+
+from config import config
+from models import Endpoint, RawFinding, Severity
+
+logger = logging.getLogger(__name__)
+
+_SSRF_PAYLOADS = [
+    "http://169.254.169.254/latest/meta-data/",
```
coolhandle01, line 19: Exactly this? Is this something I should recognise?

tools/recon/dirfuzz.py
```diff
@@ -0,0 +1,98 @@
...
+from tools._helpers import _require_binary, _run
+
+logger = logging.getLogger(__name__)
+
+# Status codes worth reporting. 403 is included because it confirms a path
+# exists and may be bypassable; 401 signals auth-gated resources.
+_INTERESTING_CODES = {200, 204, 301, 302, 307, 401, 403}
+
+
+def discover_paths(endpoints: list[Endpoint]) -> list[Endpoint]:
```
coolhandle01, line 27: Some part of this needs to be scanning the html to get all the js and CSS URLs - maybe it's like a vulnerable bootstrap or old angular or vulnerable react or something, or maybe it's a buggy custom SPA with source maps.. maybe it uses the servers old vulnerable node_modules in the front end, or maybe there's like a worker that's exploitable or.. eesh. Maybe it's in S3 or blob storage or Google storage or whatever, in which case we'll have some cloud misonfigs to check for.

You know I'm wondering if all these guys need like.. a browser xD.

tools/recon/cert_transparency.py
```diff
@@ -0,0 +1,27 @@
+"""Subdomain discovery via crt.sh certificate transparency logs."""
```
coolhandle01, line 1: On the certificate, we should like, look at that. Is it self signed, is it expiring, etc. I wonder what DNS coverage we have.

coolhandle01: Like do we care about whois and dig and stuff like that?

tools/recon/httpx.py
```diff
@@ -0,0 +1,46 @@
+"""Live host probing via the httpx CLI."""
```
coolhandle01: Do we need httpx and if so should we standardise? Could be a pain.

PR #42, opened 2026-05-09
fix: stuff that broke when I ran it plus UI (WIP)

coolhandle01:
## Summary

This fixes a bug introduced in a fix for the default model not having the anthropic/ prefix
🫠

It also optimised the PM agent by replacing a set of tools with a single tool to find valid H1 programs for consideration.

Finally, a script is introduced to install all the external dependencies for the cybersquad.

- `estimate_cost` was matching model strings like `claude-sonnet-4-6` against pricing prefixes, but the model string passed in at runtime includes the litellm provider prefix (`anthropic/claude-sonnet-4-6`), causing the lookup to always miss and return $0.00
- Fix: strip everything up to and including the first `/` before the prefix match
- Added a regression test asserting that `anthropic/claude-sonnet-4-6` and `claude-sonnet-4-6` produce identical non-zero costs

Also UI. This started as me fixing bugs that I found by actually running cybersquad.

## Test plan

- [x] `pytest tests/test_metrics.py` passes (14 tests)
- [x] Full CI stack passes locally
- [ ] Next pipeline run shows actual cost figures in SQUAD METRICS

### DIFF
--- a/install-tools.sh
+++ b/install-tools.sh
@@ -13,7 +13,7 @@
 set -euo pipefail
 
 GO_MIN_VERSION="1.22"
-GO_INSTALL_VERSION="1.23.4"  # fallback tarball version if apt is too old
+GO_INSTALL_VERSION="1.23.4"
 BIN_DIR="${HOME}/.local/bin"
 GOPATH="${GOPATH:-${HOME}/go}"
 GOBIN="${GOPATH}/bin"
--- a/install-tools.sh
+++ b/install-tools.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# install-tools.sh - Install external binaries required by the Bounty Squad pipeline.
+# install-tools.sh - Install external binaries required by the cybersquad pipeline.
 #
 # Targets Ubuntu/Debian (WSL). Run once before your first pipeline run.
 # Safe to re-run - skips tools that are already on PATH.
--- a/install-tools.sh
+++ b/install-tools.sh
@@ -136,13 +136,12 @@ nuclei -update-templates -update-template-dir "${TEMPLATES_DIR}" 2>/dev/null \
 
 # ---- PATH update ----
 
+# shellcheck disable=SC2016  # single quotes intentional: ${HOME} must expand at source-time, not now
 PATH_LINE='export PATH="${HOME}/.local/bin:${HOME}/go/bin:/usr/local/go/bin:${PATH}"'
 
 for rc_file in "${HOME}/.bashrc" "${HOME}/.zshrc"; do
   if [[ -f "${rc_file}" ]] && ! grep -qF 'go/bin' "${rc_file}"; then
-    echo "" >> "${rc_file}"
-    echo "# Added by cybersquad install-tools.sh" >> "${rc_file}"
-    echo "${PATH_LINE}" >> "${rc_file}"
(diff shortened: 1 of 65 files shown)


install-tools.sh
```diff
@@ -0,0 +1,167 @@
...
+  local bin_path
+  bin_path="$(find "${tmp}/extracted" -type f -name "${name}" | head -1)"
+  if [[ -z "${bin_path}" ]]; then
+    echo "ERROR: could not find ${name} binary in archive"
+    rm -rf "${tmp}"
+    return 1
+  fi
+
+  install -m 755 "${bin_path}" "${BIN_DIR}/${name}"
+  rm -rf "${tmp}"
```
coolhandle01: God forbid tmp be null

coolhandle01:
Things that need tidying up:

### formalisation about the program/run directory

- I think it's still fragile to give an example that implies my name is 'user' in the Programme Manager's expected outputs.
- need to make sure all the crew tools produce artifacts in the program run directory under logs/<host>. All agents should be able to choose to look at/read/parse these `<user_home>/<program>/<run_id>/logs/` artifacts, and these artifacts are useful for me to debug, and useful for H1 and their customers in discovery when they inevitably sue me for going out of scope with a robot. Honesty and evidence are paramount to this whole thing being realistically viable I think. Agents should describe the artifacts they've created as part of their expected output so the next agent knows where and what it's looking at. I do like our RawFindings and Endpoint etc data types, and we should continue to use them, but
- agents should record their findings to disk. The data that the technical author and disclosure coordinator and maybe even vulnerability researcher work with (the osint and pentest findings) should be on disk in the run directory for the most part.

### the ui looks amazing fair play

I quite like the UI. I got scared 'cos all the tasks were listed and I CTRL+Q'd out.

It looks like we're still calling `build_crew` and thus `build_tasks`, so this will respect the hack I did in tasks.py's `build_tasks` where I only return the agent's that I want to shake down.. 

image

but the panel on the left listed them all due to the use of the enum in tui.py:

```py
_TASKS = [
    "Programme Manager",
    "OSINT Analyst",
    "Penetration Tester",
    "Vulnerability Researcher",
    "Technical Author",
    "Disclosure Coordinator",
]
```

Can we factor this _TASKS away? It looks like you'll need the `crew` before `compose(self) -> ComposeResult` is called. The UI should only show tasks that it will run, it needs a bit of decoupling I think.. (you might have accidentally invented quite a nice reusable UI for Crew.AI pipelines?!).

coolhandle01:
bugs remaining

I got my PATH issues sorted out. Subfinder is working ROFL. waybackurls is working.

Hello Cloudflare 👋  🫢

I think tracepath is working?

```
[19:16:18] WARNING  Command exited 1: /usr/bin/tracepath: ztx-webhooks.one.dash.cloudflare.com: No address associated with hostname
```

I think httpx is not working, maybe I can fix this:

```
[19:14:47] WARNING  Command exited 2: Usage: httpx [OPTIONS] URL
                    Error: No such option: -s
           INFO     httpx probed 0 live endpoints from 4 hosts
```

Hmm, maybe not:
image


I think the OSINT output is too long, there should be a json path used here instead of a huge dump. I can't see all of the response in the crew.ai trace! It looks like it might be getting cut off:
image

image

Issue #33, opened 2026-05-09
feat: CSRF detection tool for penetration tester


coolhandle01:
## Summary

The penetration tester agent has no CSRF detection capability. This is a straightforward tool implementation similar to `cors.py` or `webapp_headers.py`.

## What to implement

New file: `tools/pentest/csrf.py`

Checks to perform against discovered URLs:

- Missing `SameSite` attribute on session/auth cookies (should be `Strict` or `Lax`)
- Missing `Secure` and `HttpOnly` flags on sensitive cookies
- State-changing endpoints (POST/PUT/PATCH/DELETE) that accept requests without a CSRF token header or double-submit cookie pattern
- Origin/Referer header not validated - send a cross-origin `Origin: https://evil.example.com` header and check if the request succeeds
- Forms lacking a hidden CSRF token field

## Acceptance criteria

- `@tool`-decorated function wired into `squad/penetration_tester/__init__.py`
- All network calls mocked in unit tests under `tests/test_csrf.py`
- `@pytest.mark.unit` on all tests
- Passes the full CI stack locally

PR #80, opened 2026-05-15
feat: CSRF detection tool for penetration tester (#33)

coolhandle01:
## Summary

- Adds `tools/pentest/csrf.py` with `check_csrf(endpoints)` covering two detection tiers
- Uses BeautifulSoup4 for HTML parsing (added as an explicit project dependency)
- Wires the tool into the PT agent as `csrf_check_tool` in `squad/penetration_tester/__init__.py`
- Adds `tests/test_csrf.py` with 40 unit tests covering all detection paths

## Detection tiers

**Tier 1 - Missing CSRF token in POST form (MEDIUM)**

GETs each endpoint and parses the HTML with BeautifulSoup. Before inspecting forms, checks for page-level CSRF protection signals that indicate the app protects POSTs without hidden inputs:

- A CSRF-pattern cookie in the response (Angular's `XSRF-TOKEN`, Django's `csrftoken`, Tornado's `_xsrf`) - double-submit cookie pattern
- A `<meta name="csrf-token">` tag in the HTML (Rails, Spring) - JS reads the token from the DOM and sends it as a request header

If either signal is present the endpoint is skipped entirely. Otherwise, any POST form without a hidden input whose `name` matches a CSRF-token pattern (`csrf`, `token`, `_token`, `authenticity_token`, `nonce`, `xsrf`) is flagged.

**Tier 2 - Origin header not validated (HIGH)**

For each endpoint where Tier 1 found an unprotected POST form, POSTs to the form action URL twice: once with `Origin: https://evil.example.com` and once with the correct origin. If both return the same 2xx status the server is not validating the Origin header.

One finding per endpoint per tier.

## Test plan

- [x] Detects missing CSRF token in POST form -> MEDIUM
- [x] No finding when CSRF token hidden input is present
- [x] No finding when XSRF-TOKEN / csrftoken cookie is set (Angular/Django)
- [x] No finding when `<meta name="csrf-token">` is present (Rails)
- [x] Session cookies (JSESSIONID etc.) do not suppress findings
- [x] Detects origin not validated (same 2xx for evil and correct origin) -> HIGH
- [x] No finding when origin is validated (different status codes)
- [x] Tier 2 skipped when page-level cookie protection present
- [x] Skips non-HTML responses
- [x] Skips 5xx endpoints
- [x] Network exceptions swallowed, deduplication, one finding per endpoint per tier
- [x] 605 unit tests pass, all CI checks green

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ dependencies = [
     "crewai-tools>=0.15.0",
     "pydantic>=2.7.0",
     "requests>=2.32.0",
+    "beautifulsoup4>=4.12,<5.0",
     "dnspython>=2.6",
     "python-dotenv>=1.0.0",
     "rich>=13.0",
@@ -29,6 +30,7 @@ dev = [
     "ruff>=0.4",
     "mypy>=1.10",
     "types-requests>=2.32",
+    "types-beautifulsoup4>=4.12",
 
     # SAST
     "bandit[toml]>=1.7",
--- a/squad/penetration_tester/__init__.py
+++ b/squad/penetration_tester/__init__.py
@@ -150,16 +150,19 @@ def csrf_check_tool(recon_result_json: str) -> list[dict]:
     """
     Detect CSRF vulnerabilities across HTML endpoints in the recon surface.
 
-    Tier 1 (MEDIUM): fetches each HTML endpoint and parses POST forms.  Any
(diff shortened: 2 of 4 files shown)


coolhandle01:
Tier 1 - IIRC another pattern to identify crsf tokens is hidden text/password inputs?

coolhandle01:
I wouldn't object to beautifulsoup if that's easier/nicer to work with

PR #91, opened 2026-05-16
feat: tools/html.py fetch helper, migrate csrf.py

### LINKED ISSUE #81: feat: add tools/html.py - thin HTML fetch+parse helper over http.py
## Summary

Now that `beautifulsoup4` is a first-class dependency (added in the CSRF PR), it makes sense to give HTML parsing a dedicated home rather than inlining `BeautifulSoup(resp.text, "html.parser")` at every call site.

## Proposed design

Add `tools/html.py` exposing a `fetch()` function that wraps `http.get` and returns a `(Response, BeautifulSoup)` pair:

```python
# tools/html.py
from __future__ import annotations

from bs4 import BeautifulSoup
import requests
from tools import http

def fetch(url: str, **kwargs: object) -> tuple[requests.Response, BeautifulSoup]:
    resp = http.get(url, **kwargs)
    ct = resp.headers.get("Content-Type", "")
    body = resp.text if "text/html" in ct else ""
    return resp, BeautifulSoup(body, "html.parser")
```

Call sites become:

```python
resp, soup = html.fetch(ep.url, timeout=config.recon.http_timeout, allow_redirects=True)
for form in soup.find_all("form"):
    ...
```

## Motivation

- Keeps `http.py` focused on UA injection / transport delegation only.
- Avoids a polymorphic return type (`Response` vs `(Response, ResultSet)`) if we tried to bolt `query=` onto `http.get` directly.
- Gives future tools (SRI checks, content sniffing, link extraction) a clean one-liner entry point instead of repeating the two-step get + parse pattern.
- `tools/pentest/csrf.py` is the first natural consumer; migrate it as part of this issue.

## Blocked by / related

- CSRF detection PR (which introduced `beautifulsoup4` as a dep) should land first.

coolhandle01:
Closes #81

## Summary

- **`tools/html.py`** - new `fetch(url, **kwargs) -> (Response, BeautifulSoup)` helper that wraps `http.get` and returns an empty soup for non-HTML `Content-Type`, so callers get a clean one-liner entry point rather than repeating the two-step get + parse pattern
- **`tools/pentest/csrf.py`** - first consumer migrated: `check_csrf` now calls `html.fetch` instead of `http.get` + manual content-type check; `_parse_post_forms` and `_has_csrf_meta_tag` now accept a `BeautifulSoup` object (parsing happens once at the fetch layer, not twice inside each helper)
- **`tests/test_html.py`** - new unit tests covering HTML parse, non-HTML skip, missing Content-Type header, kwargs forwarding, caller timeout respected
- **`tests/test_csrf.py`** - direct helper tests updated to construct `BeautifulSoup` via a local `_soup()` helper; integration tests unchanged (mock at `requests.get` level still intercepts through the new stack)

## Design note

`html.fetch` returns an empty `BeautifulSoup` (rather than `None`) when Content-Type is not `text/html`. This means callers can rely on soup methods returning empty results naturally, without an explicit header check — `soup.find_all("form")` just returns `[]`.

## Test plan

- [x] 675 unit tests pass (7 new)
- [x] `ruff check` + `ruff format` clean
- [x] `mypy --ignore-missing-imports` clean (100 source files, 0 errors)
- [x] `bandit` clean

### DIFF
--- a/tests/test_csrf.py
+++ b/tests/test_csrf.py
@@ -6,6 +6,7 @@ from collections.abc import Callable
 from unittest.mock import MagicMock, patch
 
 import pytest
+from bs4 import BeautifulSoup
 
 from models import Endpoint, Severity
 from tools.pentest.csrf import (
@@ -51,6 +52,10 @@ _HTML_GET_FORM_ONLY = """
 _HTML_NO_FORM = "<html><body><p>Hello world</p></body></html>"
 
 
+def _soup(html: str) -> BeautifulSoup:
+    return BeautifulSoup(html, "html.parser")
+
+
 def _post_resp(status: int = 200) -> MagicMock:
     resp = MagicMock()
     resp.status_code = status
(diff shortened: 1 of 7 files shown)


coolhandle01:
Is csrf the only candidate?

coolhandle01:
No - two others would benefit from migrating to `html.fetch()`:

**`tools/pentest/sri.py`** - fetches the page then runs `_SCRIPT_RE.finditer(html)` / `_LINK_RE.finditer(html)` to find `<script>` and `<link>` tags. Using `soup.find_all("script", src=True)` / `soup.find_all("link", rel="stylesheet")` would be cleaner and drop the two compiled regexes.

**`tools/pentest/sourcemaps.py`** - fetches the page then runs `_JS_SRC_RE.findall(page_resp.text)` to find `<script src>` tags. Same deal - `soup.find_all("script", src=True)` replaces the regex and skips the manual Content-Type check.

**`tools/pentest/errors.py`** - fetches the page but searches `resp.text[:3000]` for raw error strings (Python tracebacks, DB error messages, etc.). That is plain-text pattern matching, not HTML structure traversal, so BeautifulSoup would not add anything there.

Happy to pick up `sri.py` and `sourcemaps.py` as a follow-on issue/PR if you want to keep the scope of this one tight.

---

Issue #83, opened 2026-05-15
feat: tools/pentest/framework/ - framework-specific vulnerability probes


coolhandle01:
## Motivation

The CSRF tool already detects framework signals (Django csrftoken cookie, Rails csrf-token meta tag, Angular XSRF-TOKEN) and adjusts its behaviour accordingly.  That pattern will recur: once you know the framework, you can run targeted probes that generic tools miss.

Precedent is `tools/cloud/databases/` - when per-engine logic got heavy enough it earned its own directory.  The trigger here should be the same: a second or third framework-specific probe that would otherwise bloat an unrelated tool.

## Proposed structure

```
tools/pentest/framework/
    __init__.py
    django.py    # debug mode, admin panel, ALLOWED_HOSTS misconfiguration,
                 # @csrf_exempt endpoint detection
    rails.py     # verbose error pages leaking stack traces, secrets in
                 # /rails/info, skip_before_action bypass detection
    spring.py    # Actuator endpoints (/actuator/env, /actuator/heapdump),
                 # Spring Boot banner, csrf().disable() bypass detection
    nextjs.py    # API routes bypassing CSRF server actions, exposed
                 # /_next/static source maps (overlaps with sourcemaps.py)
```

## Framework detection

Each framework should provide a method of identifying it based on available data points observe with existing tools. Such a helper could inspect response headers (`X-Powered-By`, `X-Generator`), cookies, and HTML signals and return a `Framework` enum value.  Tools that already do this ad-hoc (CSRF, headers) could be delegated to.

## Relationship to existing issues

- Builds on the BeautifulSoup foundation added in the CSRF PR (#33)
- `tools/html.py` (#81) would be a natural complement - framework detection often requires parsing HTML
- Do not create this directory until at least two framework-specific probe files exist - one file does not justify a subdirectory
- surface detection for #76 , #75, #73, #74 and #78 etc..
  - boy scout rule applies, consider discussion around DiscoveredMCP type

PR #102, opened 2026-05-16
feat(pentest): IDOR probe tool + attack-classification skill (#63)

### LINKED ISSUE #63: feat: test for Insecure Direct Object Reference (IDOR)
## Summary

No tool currently tests for IDOR vulnerabilities. IDOR is the most frequently rewarded vulnerability class on HackerOne and the #1 category in OWASP Top 10 2021 (A01). Without it, the squad is leaving a significant chunk of the attack surface untested.

## What to implement

New file: `tools/pentest/idor.py`

### Discovery phase

- Collect all endpoints from the recon result that include numeric or UUID-style path segments or query parameters (e.g. `/users/12345`, `?order_id=abc-123`, `/api/v1/documents/42`)
- Identify parameters that look like object identifiers: `id`, `user_id`, `account_id`, `order_id`, `doc_id`, `file_id`, `invoice_id`, etc.

### Probe phase

For each candidate parameter:

- **Sequential ID probing** - if the observed value is numeric, try `value - 1`, `value + 1`, `value - 100`. A 200 response returning data that differs from the authenticated session's own data is a candidate finding.
- **Predictable UUID/slug swap** - if a second test account is available (not required), swap identifiers between sessions. If not available, flag the endpoint as manually reviewable.
- **Zero / negative IDs** - `id=0`, `id=-1`. Unexpected 200 responses with data are a finding.
- **Type juggling** - `id=1.0`, `id=1e1`, `id=01`. Some backends resolve these to the same object while bypassing access checks.

### Detection heuristics

- Response body contains personal data fields (email, name, address, payment info) that don't match the requesting session - HIGH
- Response status changes from 403 to 200 when ID is varied - HIGH
- Response body length increases significantly vs the baseline - MEDIUM (needs manual review)
- 200 response for `id=0` or `id=-1` - LOW/MEDIUM

### Scope constraint

Only probe endpoints and parameters that are explicitly in scope per `filter_in_scope()`. Do not brute-force ID ranges - limit to the small set of heuristic variants above to stay non-disruptive.

## H1 weakness types

- `Insecure Direct Object Reference (IDOR)`

## Acceptance criteria

- `@tool`-decorated function wired into `squad/penetration_tester/__init__.py`
- `weakness_types = ["Insecure Direct Object Reference (IDOR)"]` set on the function
- All network calls mocked in `tests/test_idor.py`
- `@pytest.mark.unit` on all tests
- Full CI stack passes

coolhandle01:
## Summary

- Implements `tools/pentest/idor.py` - IDOR probe tool for OWASP A01 / HackerOne's most-rewarded class (closes #63)
- Adds `cybersquad-pentest-tool` skill to enforce the attack-classification pattern on all future pentest tool work

## What the tool does

Probes numeric path segments (`/api/users/12345`) and ID-shaped query parameters (`id`, `user_id`, `order_id`, ...) using three named attack strategies via `IDORAttack` StrEnum:

- `sequential` - neighbour IDs (value+/-1, value-100); for path segments where the current value is visible in the URL
- `boundary` - zero and negative IDs (0, -1); catches missing lower-bound checks on any ID-shaped param
- `type-juggling` - loose-numeric probes (1.0, 1e1, 01); targets backends that may bypass strict access-control comparisons

Detection signals: 401/403 -> 200 transition (HIGH), PII pattern in 200 response (HIGH), unexpected 200 for boundary ID (MEDIUM). One finding per endpoint; stops on first confirming probe. No ID-range brute-forcing.

Wired into the PT squad as `idor_probe_tool` with `IDORAttack` exposed so the agent can select a targeted subset (e.g. `attacks=["boundary"]` for a quiet recon pass, 2 requests per candidate instead of many).

## What the skill enforces

`cybersquad-pentest-tool` fires on edits to `tools/pentest/` and `squad/penetration_tester/__init__.py`. It codifies three checkpoints that must hold for any multi-variant active probe tool:

1. A `StrEnum` (`ToolNameAttack` or `ToolNamePayload`) with per-member docstrings explaining when an agent picks each variant
2. `check_*` function accepts `list[XxxAttack] | None = None`; `None` = all via `frozenset(XxxAttack)`, `[]` = no-op
3. Squad `@tool` wrapper imports the StrEnum, exposes it in the typed signature, and documents each variant

Explicit exemptions (single-vector tools, binary wrappers, structural checkers) are stated with reasoning.

## Test plan

- [x] 19 unit tests in `tests/test_idor.py` - all network calls mocked, `@pytest.mark.unit`
- [x] Covers all three detection signals, all three attack strategies, filter semantics (boundary-only, type-juggling-only, empty list no-op, None runs all), skip conditions (500 endpoints, non-ID params, no numeric segment), deduplication, and exception swallowing
- [x] Full CI stack passes locally: ruff, mypy, pytest (91.27% coverage, above 90% floor), bandit

### DIFF
--- a/tools/pentest/idor.py
+++ b/tools/pentest/idor.py
@@ -1,0 +1,267 @@
+"""IDOR (Insecure Direct Object Reference) detection - probe numeric ID
+path segments and ID-shaped parameters with heuristic variants to detect
+access-control bypass and personal data exposure."""
+
+from __future__ import annotations
+
+import logging
+import re
+from enum import StrEnum
+from urllib.parse import urlparse, urlunparse
+
+import requests
+
+from config import config
+from models import Endpoint, RawFinding, Severity
+from tools import http
+from tools._helpers import adaptive_sleep
+
+logger = logging.getLogger(__name__)
+
+
+class IDORAttack(StrEnum):
+    """Named IDOR attack strategy variants.
+
+    sequential    - neighbour IDs: value-1, value+1, value-100. Only meaningful
+                    for numeric path segments where the current value is known.
+                    For query parameters (no known value) this probes 1 and 2.
+    boundary      - boundary IDs: 0 and -1. Catches missing lower-bound checks.
+    type_juggling - type-confusion probes: 1.0, 1e1, 01. Targets backends that
+                    accept loose numeric input and may resolve these to the same
+                    object while bypassing strict access-control comparisons.
+    """
+
+    sequential = "sequential"
+    boundary = "boundary"
+    type_juggling = "type-juggling"
+
+
+# Parameter names that commonly represent object identifiers.
+_ID_PARAMS = frozenset(
+    {
+        "id",
+        "user_id",
+        "account_id",
+        "order_id",
+        "doc_id",
+        "file_id",
+        "invoice_id",
+        "record_id",
+        "item_id",
+        "object_id",
+        "resource_id",
+        "customer_id",
+        "product_id",
+        "report_id",
+        "ticket_id",
+        "message_id",
+        "uuid",
+        "uid",
+        "pid",
+        "oid",
+    }
+)
+
+# Patterns that suggest a response body contains personal data.
+_PII_RE = re.compile(
+    r"[a-zA-Z0-9._%+\-]+@[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}"  # email address
+    r'|"(?:email|phone|address|payment|ssn|dob|passport)"'  # sensitive JSON keys
+    r"|card_number|credit_card|account_number",
+    re.I,
+)
+
+# A numeric path segment: at least 2 digits as a whole path component.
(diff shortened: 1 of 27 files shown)


tools/pentest/idor.py
```diff
@@ -0,0 +1,267 @@
...
+        "ticket_id",
+        "message_id",
+        "uuid",
+        "uid",
+        "pid",
+        "oid",
+    }
+)
+
+# Patterns that suggest a response body contains personal data.
```
coolhandle01: this looks like a job for #91 ..

tools/pentest/idor.py
```diff
@@ -0,0 +1,267 @@
...
+                    object while bypassing strict access-control comparisons.
+    """
+
+    sequential = "sequential"
+    boundary = "boundary"
+    type_juggling = "type-juggling"
+
+
+# Parameter names that commonly represent object identifiers.
+_ID_PARAMS = frozenset(
```
coolhandle01, line 41: this sounds like a piece of #83

Issue #121, opened 2026-05-17
test: BDD coverage for agent prompts and skills (pytest-bdd)


coolhandle01:
Follows from #26. Each squad member should be testable in isolation — mocked inputs, Gherkin-checked outputs — so that changes to prompts or skills have a clear pass/fail signal.

## Phase 1 — Spike: Programme Manager in isolation

- Introduce `pytest-bdd` and write `.feature` files to test the PM agent in isolation with mocked inputs.
- Unit tests (`-m unit`) and BDD tests (`-m bdd`) must be independently runnable.
- Goals: support skills development, support prompt fuzzing, catch reasoning regressions.
- When the pattern is working, capture it in a new repo skill (touching `docs/contributing/adding-an-agent.md` and `CLAUDE.md` as needed) so the next agent is easier to cover.
- Update CI to run BDD tests after the existing jobs. PR.

## Phase 2 — Expand to remaining agents

Onwards to recon, then the rest, one agent per PR:

- Introduce fixtures for Pydantic context inputs as needed.
- Update skill/docs after each agent.
- CI coverage for each.

## Phase 3 — DVWA sandboxed e2e (can run parallel to Phase 2)

Use pytest fixtures to stand up DVWA as a sandboxed target. Convince the PM there's a H1 programme for DVWA with scope pointing at localhost. Agents run for real against an intentionally vulnerable app.

coolhandle01:
Surfaced during review of #140 (`cybersquad-task` SKILL.md line 19, the "Prose lives in markdown, not Python strings" directory diagram).

The current layout:

```
squad/<member>/
    role.md
    goal.md              <- per-agent
    backstory.md
    <task_name>/description.md
    <task_name>/expected_output.md
```

`goal.md` is per-agent. That works for single-task members, but the Vulnerability Researcher already has two tasks (`research/` and `triage/`) and her single `goal.md` has to serve both - it ends up either generic ("find and verify") or biased toward whichever job was written first.

For BDD Phase 3 (DVWA e2e), the VR will be exercised in both modes back-to-back against a real target. A shared goal blurs the prompt's discriminative signal and makes the BDD assertions for her two tasks share noise.

**Proposed fix (for whoever picks up Phase 3):** allow `goal.md` to live inside the task subdirectory for multi-task agents, with the per-agent `goal.md` as the fallback when the per-task file is absent. `SquadMember.read()` already has the loader; this becomes one extra lookup.

[...]

PR #122, opened 2026-05-17
test: BDD scaffold — pytest-bdd infrastructure and skill

coolhandle01:
Part of #121 (M1 — infrastructure only, no BDD tests yet).

## What's here

- `pytest-bdd>=7.2` added to dev dependencies
- `bdd` marker registered in `pyproject.toml` with a description that makes the cost/speed contract explicit
- `bdd_features_base_dir = "tests/features"` set so step definitions can use short relative paths
- `tests/features/` and `tests/bdd/` directories created
- `cybersquad-bdd` skill covering: when to write BDD vs unit tests, file layout, how to run, Gherkin conventions, step definition patterns, fixture injection, and the mock contract (mock H1 API, not the LLM)
- `CLAUDE.md` index updated

## What's not here

No `.feature` files or step definitions yet — that's M2 (Programme Manager spike).

## Test plan

- `pytest -m unit` still passes (no BDD tests exist to accidentally pollute the run)
- `pytest -m bdd` returns zero tests collected, zero failures
- CI stack clean: ruff, mypy, bandit all pass

### DIFF
--- a/.claude/skills/cybersquad-bdd/SKILL.md
+++ b/.claude/skills/cybersquad-bdd/SKILL.md
@@ -42,7 +42,7 @@ pytest
 
 ## Token cost tracking
 
-CrewAI returns token usage on `CrewOutput.token_usage`. Use `estimate_cost()` from `tools/metrics.py` to convert counts to USD and log the cost at the end of each BDD scenario. This gives a running record of what each agent costs per test run and lets you spot prompt regressions that cause runaway token usage.
+CrewAI returns token usage on `CrewOutput.token_usage`. Use `estimate_cost()` from `tools/metrics.py` to convert counts to USD and log the cost at the end of each BDD scenario. This gives a running record of what each agent costs per test run and surfaces prompt regressions that cause runaway token usage.
 
 Add this to `tests/bdd/conftest.py` when writing the first BDD test:
 
--- a/.claude/skills/cybersquad-bdd/SKILL.md
+++ b/.claude/skills/cybersquad-bdd/SKILL.md
@@ -7,7 +7,7 @@ description: Gherkin conventions, step definition patterns, fixture injection, r
 
 BDD tests exercise agent *reasoning* — prompts, skills, and how an agent uses its tools given a context. They are not for testing tool logic (that is unit tests). They call a real LLM, are slow, and cost tokens.
 
-**BDD tests require a real `ANTHROPIC_API_KEY` in your local environment.** They cannot be run by Claude in a remote execution environment (no access to your key), and they are not run in CI until an API key is wired up via Secrets. Run them locally when working on prompts or skills.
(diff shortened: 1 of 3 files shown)


.claude/skills/cybersquad-bdd/SKILL.md
```diff
@@ -0,0 +1,121 @@
+---
+name: cybersquad-bdd
+description: Gherkin conventions, step definition patterns, fixture injection, and running BDD tests in isolation. Use when creating or editing files under tests/features/ or tests/bdd/.
+---
+
+# cybersquad BDD tests
+
+BDD tests exercise agent *reasoning* — prompts, skills, and how an agent uses its tools given a context. They are not for testing tool logic (that is unit tests). They hit a real LLM, are slow, and cost tokens. Run them intentionally, not as part of every commit.
```
coolhandle01: unlikely Claude wlll be able to run these with my Anthropic API key.. one day, the pipeline will contain my Anthropic API key via GitHub Secrets or something, but only when this sha1-halud malarkey is satisfactory mitigated. At the moment it just feels too risky..

.claude/skills/cybersquad-bdd/SKILL.md
```diff
@@ -0,0 +1,121 @@
...
+## Running
+
+```bash
+# BDD tests only
+pytest -m bdd
+
+# Unit tests only (default CI run - does not include BDD)
+pytest -m unit
+
+# Both
```
coolhandle01: whaddabout just `pytest` ?

.claude/skills/cybersquad-bdd/SKILL.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+@then("the selected programme allows automated scanning")
+def check_allows_scanning(pm_output):
+    assert pm_output.allows_automated_scanning is True
+```
+
+## Fixture injection rules
+
+- Use `target_fixture` on `@given` and `@when` steps to name the fixture the step produces.
```
coolhandle01, line 138: Maybe I'm dreaming but, can we add a fixture for token costs of the tests, and some logging to print that out? One day, I'll wonder "how much did the repo spent on BDD tokens this month?".

Maybe we can reuse the existing metrics stuff, possibly we don't even need a fixture.

I think I'd like to surface a significant change in token usage, certainly token cost of the agent on it's task should be a test signal, possibly being considered a regression at some threshold for a spike. Not necessarily a problem for Gherkin, of course.

pyproject.toml
```diff
@@ -56,7 +57,9 @@ addopts = [
 markers = [
     "unit: fast, no I/O, no external calls",
     "integration: hits real APIs or binaries (skipped in CI by default)",
```
coolhandle01, line 59: Err.. yeah integration tests.. that's a good idea. I should do some of those one day lol. I hadn't considered that this marker was already defined when I was thinking about bdd for the agents, .. TBD

PR #123, opened 2026-05-18
feat(ta): make the Technical Author actually author

coolhandle01:
"In a vacuum" PR for #114, per your request. Branched off `feat/file-path-artifacts` so you can review the TA tooling idea on its own and either pull it in, cherry-pick parts, or close it without affecting #114.

## The problem I tried to solve

Reading `squad/technical_author/write/description.md` next to `tools/report_tools.py`, the TA's prompt and tooling don't agree on what the TA's *job* is.

The prompt asks for craft: a punchy title that matches `[Type] in [Component] allows [Outcome]`, a developer-tuned description that explains **why** the code is vulnerable, sanitised evidence with no live credentials, a specific impact statement, and remediation that cites OWASP or CWE.

The only tool that writes a report is `create_reports_tool(verified_path, programme_handle, summary)`. It reads `VerifiedVulnerability` from `verified.json` and pours every field — `title`, `description`, `steps_to_reproduce`, `evidence`, `impact`, `remediation` — directly into a markdown template. The TA's only inputs are the programme handle and a 2-3 sentence session summary; every other byte comes from the VR. The TA can't be punchy, developer-focused, sanitised, or citation-rich, because there's no argument those words could even flow through.

Three smaller things compounded that:

1. Raw `evidence` (sqlmap output, full request/response dumps, JWTs) was inlined verbatim up to 2000 chars. A bearer token or session cookie in the VR's evidence shipped straight to H1.
2. CWE lookup was a 13-entry hardcoded `dict` in `report_tools.py`. It listed `XSS` but not `ReflectedXSS`/`HeaderXSS`, listed `IDOR` but not `AccessControlBypass`, and missed `PrototypePollution`, `SSTI`, `OpenRedirect`'s several variants, JWT, etc. — many of the actual `vuln_class` strings the pentest tooling emits.
3. No quality gate. The "triage on first read" bar in the prompt isn't enforceable when nothing checks it.

## What's in the PR

Six new tools for the TA, one removed, plus the supporting infrastructure.

| Tool | Role |
|---|---|
| **Sanitise Evidence** | Redacts credentials, cookies, bearer tokens, JWTs, AWS keys, secret-shaped `key=value` pairs. Preserves payloads (XSS strings, SQLi vectors, SSRF URLs) — this is private disclosure and the triager needs the literal exploit. Returns `{sanitised, redactions, warnings}`. |
[...]

PR #114, opened 2026-05-17
fix: reliable pipeline — file-path artifacts, typed contracts, full-sweep disclosure

coolhandle01:
This PR addresses several absolute stinker design issues that Claude introduced early on that are finally problematic.

#### problems

- run_recon returns a list of pydantic types that blow away most of the context window, and this was just brazenly printed to the output, which overflowed, handing the pentester half a piece of json as "speech"
- the pipeline ordering was wrong, _for ages_. I was ignored a _lot_ when I pushed back on it*.
- a single issue would be submitted after entire recon and pentest stage, an insane choice to limit RoI
- none of these agents are ready for _bdd_ (#121's M2)

*Quite annoyed about that. Lots and _lots_ of wasted cycles on AI's BS assertions, now many more will come undoing it. Hey Mississipi I hope you like data-centres poisoning the water you use for baby formula, because Claude clearly DGAF if you don't, sorry

#### fixes

- a 'workspace' for the squad to share files. It's rough. It doesn't matter yet. One day this recon stuff will all be amass and tools will query `amass db` backed by postgres. By that time, the workspace will just be about persisting evidence for the final report
  - rewrite all the prompts so agents don't spit MB of json at each other.
  - rewrite all the prompts so agents do agenting and aren't just really pricey ways to call functions
- move the VR in the pipeline: VR now briefs PT based on OSINT with an attack plan 
- ...

PR #125, opened 2026-05-18
feat(oa): make the OSINT Analyst actually curate the attack surface

coolhandle01:
"In a vacuum" PR for #124, completing the pre-PT trilogy. Branched off `claude/improve-vr-triage-tools` so you can review the OA recon-curation tooling on its own and stack the three (#125 → #124 → #123 → #114) or pick whichever you want to land.

## The problem I tried to solve

Same disease the TA and VR had before this top-down pass. Stepping into the OA's role:

I am asked for "a comprehensive, in-scope attack surface map" - subdomains, endpoints, ports, tech. What I actually have is **one** tool, `Run Recon`, that does subfinder + httpx + nmap + ffuf + TLS + DNS checks in a fixed sequence and dumps everything into `recon.json`. After that, I read it back via the query tools and write a briefing **from memory** of what I saw.

The agent has zero curation role in the *artefact*. The briefing prose is gone the moment it's consumed. Every subdomain in `recon.json` has equal weight. No host carries a role (admin vs API vs CDN vs static). No priority signal (where should the PT spend probe budget?). `ReconResult.notes` is an autogenerated `f"Seeded from [...]. {len} in-scope hosts."` literal. The tech list is a flat dedup'd `list[str]` with no per-host attribution. So the downstream consumers (VR research, PT) get an undifferentiated soup and have to do the curation work themselves *from raw inventory*, every time.

That's exactly the bug shape we just fixed for TA and VR. Apply the same pattern.

## What's in the PR

Split sweep from authoring. The OA now drives a discovery + curation loop, and the artefact persists the curation.

| Tool | Role |
|---|---|
| **Run Initial Sweep** | Renamed from Run Recon. Same subfinder/httpx/nmap/ffuf/TLS/DNS pipeline, but writes to `sweep.json` (OA-internal draft), not `recon.json`. |
| **Recon Subdomains / Recon Endpoints / Recon Open Ports** | Typed-query tools shared with PT, so the OA can inspect its own sweep before annotating. |
| **Certificate Transparency Lookup / Historical URL Discovery / LLM Endpoint Detection** | Kept - they were already there, the agent just had no reason to use them in a single-mega-tool world. |
| **Probe Hostnames** | New. Re-probes new hostnames discovered via CT / wayback. Scope-filters inputs *before* any HTTP traffic - we never probe outside scope, even for fingerprinting. |
| **Annotate Host** | New, the authoring tool. Author one `HostInsight` per interesting host: role (admin/api/auth/app/cdn/static/mail/infra/dev/unknown), priority (high/medium/low/skip), notes, detected_tech (with versions where possible). Returns a validation report. |
| **Uncovered Hosts** | New. Lists interesting-status hostnames in the sweep that have no insight yet - a checklist before `Finalise Recon`. |
| **Lookup CWE / Lookup OWASP Guidance** | Shared with TA / VR. Useful when annotating a host to ground notes in a known weakness class ("WordPress 5.8 -> CWE-79 stored XSS"). |
| **Finalise Recon** | New. Consolidates sweep + every authored `HostInsight` into the canonical `recon.json`. Refuses on hard errors. |

## The validation gate

`Annotate Host` runs `validate_insight` on every call and `Finalise Recon` re-runs it before consolidating. Errors block, warnings advise.

- **hostname** must be in programme scope (the gate has access to the structured scope and rejects out-of-scope annotations).
- **notes** >= 30 chars; >= 60 if priority is HIGH (forces a substantive reason for spending PT budget); >= 30 if priority is SKIP (so a future reader doesn't silently re-include it).
- **detected_tech** entries must be non-trivial strings.
- **Warning** when the sweep saw a tech the agent's annotation drops - signals the agent may have ignored a useful tech signal.

`Finalise Recon` additional rules:
- At least one insight authored at all.
- At least one HIGH priority host on a non-empty surface - the PT needs a focus target.
- No insight has hard errors.

## Model changes

`HostRole` and `HostPriority` StrEnums + `HostInsight` BaseModel added to `models.py`. `ReconResult` gains a backward-compatible `host_insights: list[HostInsight]` field (empty on the OA's draft `sweep.json`, populated on the final `recon.json`). The canonical filename downstream consumers read remains `recon.json` - no PT, VR, or TA change is required.

## Authoring loop

```
Run Initial Sweep(handle) -> "sweep.json"

# inspect what's there
Recon Endpoints("sweep.json", status=200)
Recon Open Ports("sweep.json")

# (optional) widen via supplementary discovery
Certificate Transparency Lookup("example.com")
Probe Hostnames([new_hosts], handle)  # scope-filtered
LLM Endpoint Detection(endpoints_json)

# annotate the interesting hosts
for each host:
    Lookup CWE / OWASP for the host's tech (optional grounding)
    Annotate Host(host, role, priority, notes, detected_tech) -> validation
    # repeat with fixes until validation.ok

Uncovered Hosts("sweep.json")  # checklist before finalisation

Finalise Recon(handle) -> "recon.json"
```

Same shape as #123 and #124. Same critique-and-iterate dynamic.

## CI

- `ruff check`, `ruff format --check`, `mypy --ignore-missing-imports`, `bandit` all clean.
- `pytest -m unit --cov` - **878 passed, 92.11% coverage** (above the 90% floor). `tools/recon_insights.py` lands at 99%.
- New file `tests/test_recon_insights.py` covering validation rules (one per rule), persistence with per-host filenames, the coverage helper, and finalisation including the "missing sweep / missing insights / no HIGH host / validation errors" guards.
- New OA tool-wrapper tests in `test_squad_tools.py` for `Run Initial Sweep`, `Annotate Host` (clean + failing), `Uncovered Hosts`, `Probe Hostnames` (scope-filter check), `Finalise Recon` (clean + missing-insights), `Lookup CWE / OWASP`.
- The existing `test_recon_tool` was renamed to `test_run_initial_sweep_tool` and asserts the new `sweep.json` filename.

## Things I deliberately did not do

- **Did not touch the PT, VR, TA, or DC.** No model breakage. `recon.json` filename and schema are preserved.
- **Did not change the recon binaries pipeline.** `run_recon()` in `tools/recon/__init__.py` is unchanged; only its caller-side write target moves from `recon.json` to `sweep.json`.
- **Did not add a passive-findings authoring loop.** The OA's `passive_findings` (TLS / DNS check output) pass through unchanged - they are orphaned today (no VR reads them on the triage path) and that's a separate ticket. Mentioned as follow-up in the comments below.
- **Did not enforce "every interesting-status host must have an insight".** Too rigid on big surfaces. `Uncovered Hosts` returns the list so the agent can be deliberate, but only the HIGH-priority floor is enforced.

## How to read this

Start with `squad/osint_analyst/recon/description.md` (the new workflow), then `tools/recon_insights.py` (the primitives - `validate_insight` is the heart), then `squad/osint_analyst/__init__.py` (the `@tool` wiring). The model additions in `models.py` (`HostRole`, `HostPriority`, `HostInsight`) are small and worth a glance for the shape.

---

I've followed this PR with a comment about a couple of recon-stack additions to `install-tools.sh` that would meaningfully widen what the OA can do downstream. Nothing blocking for this PR.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -18,7 +18,7 @@ cybersquad is a six-agent CrewAI pipeline that autonomously selects HackerOne bu
 | `squad/<member>/{description,expected_output}.md` | Two single-purpose files driving the Task description and expected output. |
 | `squad/<member>/__init__.py` | Tool functions (`@tool`) + a module-level `MEMBER = SquadMember(...)` constant. |
 | `tools/h1_api.py` | HackerOne REST client. Singleton: `from tools.h1_api import h1`. |
-| `tools/recon/` | Recon tools: subfinder, httpx, nmap, TLS, DNS, dirfuzz, waybackurls, cert transparency, traceroute, scope guard. |
+| `tools/recon/` | Recon tools: subfinder, httpx, dnsx (subdomain-takeover detection), nmap, TLS, DNS, dirfuzz, waybackurls, cert transparency, traceroute, scope guard. |
 | `tools/recon/scope.py` | `filter_in_scope()` - hard scope enforcement boundary. Do not weaken. |
 | `tools/recon_insights.py` | Host-annotation authoring primitives for the OSINT Analyst: HostInsight persistence, quality validation, and final consolidation of sweep.json + insights into recon.json. |
 | `tools/pentest/` | Pentest tools: nuclei, sqlmap, CORS, SSRF, XSS, SRI, header injection, source maps, error disclosure. |
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -20,6 +20,7 @@ cybersquad is a six-agent CrewAI pipeline that autonomously selects HackerOne bu
 | `tools/h1_api.py` | HackerOne REST client. Singleton: `from tools.h1_api import h1`. |
 | `tools/recon/` | Recon tools: subfinder, httpx, nmap, TLS, DNS, dirfuzz, waybackurls, cert transparency, traceroute, scope guard. |
 | `tools/recon/scope.py` | `filter_in_scope()` - hard scope enforcement boundary. Do not weaken. |
+| `tools/recon_insights.py` | Host-annotation authoring primitives for the OSINT Analyst: HostInsight persistence, quality validation, and final consolidation of sweep.json + insights into recon.json. |
 | `tools/pentest/` | Pentest tools: nuclei, sqlmap, CORS, SSRF, XSS, SRI, header injection, source maps, error disclosure. |
 | `tools/cloud/` | Cloud/service checks: S3, Azure Blob, per-engine databases, admin panels, branded panels. |
 | `tools/cloud/databases/` | Per-engine unauthenticated database checks: one file per engine. |
--- a/install-tools.sh
+++ b/install-tools.sh
@@ -6,7 +6,7 @@
 #
 # Tools installed:
 #   Go (if absent or < 1.22): official tarball from go.dev
-#   go install: subfinder, httpx, nuclei, ffuf, waybackurls
+#   go install: subfinder, httpx, nuclei, ffuf, waybackurls, dnsx
 #   apt         : nmap, sqlmap
 #   git clone   : testssl.sh -> /usr/local/bin/testssl.sh
 
@@ -90,6 +90,7 @@ declare -A GO_TOOLS=(
   [nuclei]="github.com/projectdiscovery/nuclei/v3/cmd/nuclei"
   [ffuf]="github.com/ffuf/ffuf/v2@latest"
   [waybackurls]="github.com/tomnomnom/waybackurls"
+  [dnsx]="github.com/projectdiscovery/dnsx/cmd/dnsx"
 )
 
 for tool in "${!GO_TOOLS[@]}"; do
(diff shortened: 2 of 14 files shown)


coolhandle01:
Two recon-stack additions that aren't in `install-tools.sh` today, both of which would meaningfully widen what the OA can do downstream:

1. **`dnsx`** (`github.com/projectdiscovery/dnsx`) — bulk DNS resolution + CNAME walking. Lets the OA flag subdomain-takeover candidates (CNAME pointing to an abandoned cloud resource like a deprovisioned S3 bucket or Heroku app) directly during annotation, which is a high-signal class that the current sweep misses entirely. Maps cleanly onto a new `Detect Takeover Candidates` tool or just enriches `Annotate Host` notes.

2. **`amass`** (`github.com/owasp-amass/amass`) — the #114 PR description already nods at it ("One day this recon stuff will all be amass…"). Subfinder alone has decent passive coverage but amass's data-source set is wider, and the OA's new curation layer has more value the broader the raw inventory is. Could be added as a second pass that feeds the same `sweep.json` pipeline.

Separately, a follow-up I noticed while doing this PR but did not address: `ReconResult.passive_findings` (TLS + DNS check output the OA collects automatically) is **orphaned on the triage path** — VR triage reads `findings.json` (PT output) and never sees passive findings. They exist in `recon.json` but nothing converts them to verified entries. Probably worth a separate ticket; the OA's authoring layer is the natural place to fold them in, but pulling that into this PR would have ballooned the diff.

---

coolhandle01:
> a follow-up I noticed while doing this PR but did not address: ReconResult.passive_findings (TLS + DNS check output the OA collects automatically) is orphaned on the triage path — VR triage reads findings.json (PT output) and never sees passive findings. They exist in recon.json but nothing converts them to verified entries. Probably worth a separate ticket; the OA's authoring layer is the natural place to fold them in, but pulling that into this PR would have ballooned the diff.

I think I need a bit more information/opinion. If my understanding is correct, the desired flow of the passive recon should be something like:
The DNS/TLS info from OA's recon should make it to the VR for their research, which the VR can use to inform attack plan laid out for the PT. Then, the VR should see it again at the triage stage, and so on. If a cert expired or a CNAME is dangling, it could be a foothold into building out something more severe.

Issue #143, opened 2026-05-21
refactor(pentest): explicit args_schema on every probe tool


coolhandle01:
## Motivation

Every pentest probe currently relies on CrewAI's signature-inference path to build the JSON schema the LLM sees when picking a tool. That path:

- crams every parameter description into the function docstring, where field-level descriptions cannot be addressed individually
- forces flat positional parameters (see the `# pylint: disable=R0913,R0917` block in `squad/programme_manager/__init__.py` lines 16-28 - the same shape will spread across pentest probes)
- gives us no place to attach validation that is richer than the type hint

For internal tooling this is fine. For **pentest probes specifically** the cost of a mis-call is real: probes hit live services on live programmes, mis-targeting costs money, generates noise the operator has to apologise for, and at the limit risks the squad's relationship with the H1 programme. The schema is a *contract*, not a hint, and the contract should be enforced upfront. "If you find a tool the agent is mis-calling, give it an explicit schema" was the wrong framing - "never let the agent mis-call in the first place" is the right one.

## What changes

Every `@tool`-wrapped function under `tools/pentest/**` and `squad/penetration_tester/__init__.py` gets an explicit Pydantic `args_schema`:

```python
from pydantic import BaseModel, Field

class IDORArgs(BaseModel):
    target: str = Field(description="Endpoint to probe, including the object id under test")
    check: IDORCheck = Field(description="Which IDOR variant to attempt")

@tool("Probe IDOR", args_schema=IDORArgs)
def idor_tool(target: str, check: IDORCheck) -> ...:
    ...
```

Properties of the refactor:

- The StrEnum-of-variants (`check_X` per the `cybersquad-pentest-tool` skill) stays - it is doing real work constraining the LLM's choice. `args_schema` does not replace it.
- Every field gets a `Field(description=...)`. The inferred path could not address fields individually; the explicit path can.
- The function signature still matches the schema field-for-field. CrewAI's runtime validates the LLM's call against the schema before invoking the function, so any drift between the two is caught.
- Underlying probe behaviour is untouched. This is a wrapper-layer refactor.

## Constraints

- **No behaviour changes.** Probes return the same shape; the `@pentest_tool` OWASP categorisation layer is preserved.
- **Skill update required.** The `cybersquad-pentest-tool` contributor skill currently documents the StrEnum + `@pentest_tool` ceremony; it gains a third element: every probe carries an `args_schema`. Skill text updates land in the same PR.
[...]

PR #145, opened 2026-05-22
refactor(squad): typed pydantic returns for every @tool wrapper (#139)

### LINKED ISSUE #139: refactor: @tool wrappers return typed pydantic models, not list[dict]
## Context

PR #138 fixed one instance of a wider pattern: a CrewAI `@tool`-wrapped function had no typed reader for the workspace artifact the VR persists, so consumers were stuck reading raw JSON bytes via `Read Run File` instead of working against the `AttackPlan` schema.

Audit of the 87 `@tool` / `@pentest_tool` / `@research_brief_tool` functions in `squad/`:

```
72  return a dict-shaped type (list[dict] or dict)
 1  returns a bare `list` (no inner type at all)
14  already typed (EndpointPage, list[str], str, float)
```

`list[dict]` means CrewAI hands the agent a JSON-serialised dict it has to reconstitute by reading the tool's docstring. The pydantic models exist (every `check_X` returns `list[RawFinding]` internally), the wrapper just `.model_dump()`s the schema away on the way out.

## The rule (paired with the `cybersquad-tool` skill)

Every `@tool`-wrapped function returns a pydantic `BaseModel` subclass or `list[BaseModel]`. Never `dict`, `list[dict]`, or bare `list`. The exception is workspace-handle strings ("findings.json" etc.) returned by `finalise_X` / `save_X` tools - those are intentional and stay as `str`.

## Migration backlog

One sweep PR is fine - they're all mechanical:
- replace `[f.model_dump() for f in check_X(...)]` with `list(check_X(...))`
- update the return annotation
- ensure the underlying model is importable in the agent file

### squad/penetration_tester/__init__.py (44 tools)

```
:114  Nuclei Scan                                    list[dict] -> list[RawFinding]
:139  SQLMap Injection Scan                          list[dict] -> list[RawFinding]
:157  Cookie Security Check                          list[dict] -> list[RawFinding]
:176  CORS Misconfiguration Check                    list[dict] -> list[RawFinding]
:188  CSRF Detection                                 list[dict] -> list[RawFinding]
:218  SSRF Probe                                     list[dict] -> list[RawFinding]
:241  Header Injection Check                         list[dict] -> list[RawFinding]
:252  Host Header Attack Check                       list[dict] -> list[RawFinding]
:263  Header XSS Probe                               list[dict] -> list[RawFinding]
:299  JS Source Map Scan                             list[dict] -> list[RawFinding]
:311  Path Traversal Probe                           list[dict] -> list[RawFinding]
:346  HTTP Parameter Pollution Probe                 list[dict] -> list[RawFinding]
:374  Server-Side Template Injection Probe           list[dict] -> list[RawFinding]
:407  Open Redirect Probe                            list[dict] -> list[RawFinding]
:436  Reflected XSS Probe                            list[dict] -> list[RawFinding]
:452  Subresource Integrity Check                    list[dict] -> list[RawFinding]
:463  Error and Stack Trace Disclosure Check         list[dict] -> list[RawFinding]
:480  S3 Bucket Check                                list[dict] -> list[RawFinding]
:492  Azure Blob Storage Check                       list[dict] -> list[RawFinding]
:504  Unauthenticated Elasticsearch Check            list[dict] -> list[RawFinding]
:520  Unauthenticated CouchDB Check                  list[dict] -> list[RawFinding]
:536  Unauthenticated Redis Check                    list[dict] -> list[RawFinding]
:552  Unauthenticated MongoDB Check                  list[dict] -> list[RawFinding]
:569  Exposed PostgreSQL Check                       list[dict] -> list[RawFinding]
:586  Exposed MySQL/MariaDB Check                    list[dict] -> list[RawFinding]
:603  Sensitive Files Check                          list[dict] -> list[RawFinding]
:619  Admin Panels Check                             list[dict] -> list[RawFinding]
:633  cPanel/WHM Check                               list[dict] -> list[RawFinding]
:646  Plesk Check                                    list[dict] -> list[RawFinding]
:658  DirectAdmin Check                              list[dict] -> list[RawFinding]
:669  Webmin Check                                   list[dict] -> list[RawFinding]
:680  Grafana Check                                  list[dict] -> list[RawFinding]
:693  Kibana Check                                   list[dict] -> list[RawFinding]
:706  Portainer Check                                list[dict] -> list[RawFinding]
:719  Consul/Vault Check                             list[dict] -> list[RawFinding]
:732  Prompt Injection Probe                         list[dict] -> list[RawFinding]
:763  NoSQL Injection Scan                           list[dict] -> list[RawFinding]
:781  LDAP Injection Probe                           list[dict] -> list[RawFinding]
:815  Command Injection Probe                        list[dict] -> list[RawFinding]
:847  XXE Probe                                      list[dict] -> list[RawFinding]
:883  Prototype Pollution Check                      list[dict] -> list[RawFinding]
:925  IDOR Probe                                     list[dict] -> list[RawFinding]
:967  JWT Vulnerability Check                        list[dict] -> list[RawFinding]
:1046 Recon Open Ports                               dict       -> needs an OpenPortsMap model (host -> list[int])
```

### squad/vulnerability_researcher/__init__.py (8 tools)

```
:81   NVD CVE Lookup                                 list[dict] -> list[CveEntry] (new model)
:92   List Programme Reports                         list[dict] -> list[ProgrammeReportSummary] (new model)
:167  List Raw Findings                              list[dict] -> list[RawFindingSummary] (new model)
:190  Read Raw Finding                               dict       -> RawFinding
:214  Lookup CWE                                     list[dict] -> list[CweEntry] (model already exists in tools.cwe_data)
:234  Lookup OWASP Guidance                          list[dict] -> list[OwaspEntry] (model already exists in tools.owasp_data)
:254  Assess Raw Finding                             dict       -> AssessmentResult (new model wrapping path + validation)
:323  Discard Finding                                dict       -> DiscardResult (new model wrapping path)
```

### squad/programme_manager/__init__.py (2 tools)

```
:15   Browse HackerOne Programmes                    list[dict] -> list[ProgrammeBrief] (or list[Programme])
:62   Hydrate HackerOne Programme                    dict       -> Programme
```

### squad/osint_analyst/__init__.py (8 tools)

```
:66   Recon Subdomains                               list       -> list[str]  (just fix the bare list)
:77   Recon Endpoints                                dict       -> reuse EndpointPage from models
:103  Recon Open Ports                               dict       -> needs OpenPortsMap model
:134  LLM Endpoint Detection                         list[dict] -> list[LlmEndpoint] (new)
:148  Probe Hostnames                                list[dict] -> list[HostProbeResult] (new)
:172  Detect Takeover Candidates                     list[dict] -> list[TakeoverCandidate] (new)
:201  Lookup CWE                                     list[dict] -> list[CweEntry]
:223  Lookup OWASP Guidance                          list[dict] -> list[OwaspEntry]
:242  Annotate Host                                  dict       -> HostAnnotation (new)
```

### squad/technical_author/__init__.py (5 tools)

```
:25   Sanitise Evidence                              dict       -> SanitisedEvidence (new)
:39   Lookup CWE                                     list[dict] -> list[CweEntry]
:61   Lookup OWASP Guidance                          list[dict] -> list[OwaspEntry]
:93   List Programme Reports                         list[dict] -> list[ProgrammeReportSummary]
:115  Draft Vulnerability Report                     dict       -> DisclosureReport (model exists)
```

### squad/disclosure_coordinator/__init__.py (2 tools)

```
:15   Submit Report                                  dict       -> SubmissionResult (new)
:27   Check H1 Duplicate                             list[dict] -> list[ReportSummary] (probably reuses ProgrammeReportSummary)
```

### squad/workspace_tools.py (3 tools)

```
:11   List Run Files                                 list[dict] -> list[RunFile] (new)
:20   Read Run File                                  dict       -> RunFileContent (new)
:29   Read Attack Plan                               dict       -> AttackPlan  (PR #138 introduced this one shaped as dict; this issue rolls it into the sweep)
```

## Sequencing

Pre-requisite (separate PR, see the `cybersquad-tool` skill work):

1. Define a `CrewAITool` Protocol in `squad/__init__.py` and retype `SquadMember.tools: list[CrewAITool]` instead of `list[Any]`.
2. Add a contract test in `tests/` that walks every `MEMBER.tools` and asserts the return annotation is a `BaseModel` subclass or `list[BaseModel]` (with a whitelist for the workspace-handle `str` returns from `finalise_X`/`save_X`).
3. The contract test will go red - that failure list is exactly this issue's backlog.

Then the sweep PR(s) here turn the test green by retyping the wrappers. Reasonable to split per agent (6 small PRs) or do it in one (~73-tool churn but mechanical).

## Acceptance

- Every `@tool`-wrapped function returns a pydantic model or `list[<Model>]`.
- The contract test passes.
- No `.model_dump()` in any wrapper body (the conversion happens inside CrewAI on serialisation).
- New models live alongside the existing ones in `models/` so they're reusable.

coolhandle01:
## Summary

Closes #139. Typed pydantic returns for every `@tool` wrapper across the squad, plus the typed-parameter follow-on that the same audit unblocked (`list[Endpoint]` replacing `endpoints_json: str` on PT). What started as a tiny `cached_tool` helper grew into the full #139 sweep when the contract test went red on the first run.

Net-zero on `cached_tool`: added in `c3f65dd`, removed in `fd11508` after [review feedback](https://github.com/coolhandle01/cybersquad/pull/145#issuecomment-4520396152) showed CrewAI 1.14.5 already memoises every `@tool` call at the agent-runtime layer.

### Typed pydantic returns (`4b67ffe`, closes #139)

Every `@tool` / `@pentest_tool` / `@research_brief_tool` wrapper in `squad/` now returns a pydantic `BaseModel` subclass or `list[BaseModel]`. The pattern this replaces - `[f.model_dump() for f in check_X(...)]` - was stripping schemas on the way out and forcing the agent to reconstitute shape from docstrings. Roughly 73 wrappers retyped; the underlying probe / lookup helpers already returned typed models, the wrappers just had to stop flattening them.

Three pieces of enforcement landed alongside the sweep so the contract is now checked in CI rather than implicit:

- **`CrewAITool` Protocol** in `squad/__init__.py` carrying `name` / `description` / `func`. Per-agent `_PentestTool` and `_ResearchBriefTool` Protocols extend it. `SquadMember.tools` is now `list[CrewAITool]` instead of `list[Any]`.
- **`tests/test_squad_tool_contracts.py`** walks every `MEMBER.tools` entry and asserts the return annotation is one of: pydantic `BaseModel` subclass, `list[BaseModel]`, a primitive (`float` for CVSS), `list[str]` for flat handle lists, or a whitelisted workspace-handle `str` (`finalise_X` / `save_X`).
- **New models grouped with their existing neighbours** in `models/`: `ProgrammeReportSummary`, `OpenPortsMap`, `LlmEndpoint`, `HostInsight` / `HostRole` / `HostPriority`, `RawFindingSummary`, `AssessmentResult`, `DiscardResult`, `SanitisationReport`, `ReportDraftResult`, plus `CveEntry` / `CWEEntry` / `OWASPEntry` lifted from their `tools/` modules.

Exception preserved: workspace-handle strings (`"findings.json"`, `"attack_plan.json"`, `"verified.json"`, the reports manifest filename) stay as bare `str` returns - those are typed handles the next agent passes to a typed reader.

### Typed endpoints parameter (`c969a96`)

Sweeps every PT wrapper that previously took `endpoints_json: str` to `endpoints: list[Endpoint]` (and `tech_tags: list[str] | None` on the nuclei wrapper, `findings: list[RawFinding]` on save_findings). Modern CrewAI's `args_schema.model_validate(...).model_dump()` accepts `list[<Model>]` directly, so the agent now sees a typed `Endpoint` schema with `$defs/Endpoint` instead of "pass a JSON string".

`_parse_endpoints` keeps its name but is now the both-shape adapter - CrewAI's runtime hands the wrapper `list[dict]` after schema validation, while direct test invocations pass `list[Endpoint]`. Pydantic's `model_validate` accepts either, so one helper covers both paths.

Skill markdown (`cybersquad-tool`, `cybersquad-pentest-tool`) updated to reflect the retired pattern.

### `cached_tool` added then dropped (`c3f65dd` + `fd11508`)

Originally added a `functools.cache`-wrapping helper for the CWE / OWASP lookup tools on TA / OSINT / VR. Reviewer pointed out that CrewAI 1.14.5 already short-circuits the second call through `tools_handler.cache.read(...)` at `tool_usage.py:274-285` before `tool.func` runs at all, so the inner cache saved no work on the agent path. The memoisation test only passed because it called `tool.func` directly, asserting an identity property (`first is second`) that CrewAI's outer cache does not preserve either way. Reverted in full, including the test and the now-stale `lookup_*.func.cache_clear()` setup calls in the OSINT / VR test suites.

Full rationale: https://github.com/coolhandle01/cybersquad/pull/145#issuecomment-4520952756

## Relationship to #143

#143 (explicit `args_schema=` Pydantic class with per-field `Field(description=...)` on every pentest probe) is **independent and unblocked** by this work, not addressed by it. The typed-parameter sweep in `c969a96` makes #143 cheaper - the probe signatures are now properly typed, so the work in #143 is "lift each signature into a BaseModel and add Field descriptions" rather than "type the params and then..." `grep "args_schema" tools/pentest/ squad/penetration_tester/` finds zero `args_schema=` arguments in this PR; #143's acceptance criteria are all still open.

## Test plan

- [x] `ruff check .` - clean
- [x] `pylint .` - 9.87/10 (no new findings vs base)
- [x] `pytest -m unit --cov --cov-report=term-missing` - 1064 passed, 91.5% coverage
- [x] Contract test (`tests/test_squad_tool_contracts.py`) green - the enforcement that closes #139

### DIFF
--- a/.claude/skills/cybersquad-pentest-tool/SKILL.md
+++ b/.claude/skills/cybersquad-pentest-tool/SKILL.md
@@ -80,10 +80,10 @@ In `squad/penetration_tester/__init__.py`, use `@pentest_tool` (not bare `@tool`
 ```python
 @pentest_tool("SSRF Probe", check_fn=check_ssrf)
 def ssrf_probe_tool(
-    endpoints_json: str,
+    endpoints: list[Endpoint],
     payloads: list[SsrfPayload] | None = None,
 ) -> list[RawFinding]:
-    return list(check_ssrf(_parse_endpoints(endpoints_json), payloads))
+    return list(check_ssrf(_parse_endpoints(endpoints), payloads))
 ```
 
 4. Document each variant in the docstring so the agent knows when to pick a subset. Include what the variant targets, when it is useful, and the request-cost implication of narrowing.
--- a/.claude/skills/cybersquad-tool/SKILL.md
+++ b/.claude/skills/cybersquad-tool/SKILL.md
@@ -16,8 +16,8 @@ def read_attack_plan_tool() -> AttackPlan:
     return load_attack_plan(attack_plan_path())
 
 @tool("Reflected XSS Probe")
-def xss_probe_tool(endpoints_json: str) -> list[RawFinding]:
-    return list(check_reflected_xss(_parse_endpoints(endpoints_json)))
+def xss_probe_tool(endpoints: list[Endpoint]) -> list[RawFinding]:
+    return list(check_reflected_xss(_parse_endpoints(endpoints)))
 ```
 
 ```python
@@ -40,7 +40,7 @@ Never `dict`, `list[dict]`, or bare `list` (no inner type). CrewAI serialises py
 # correct
 @pentest_tool("SSRF Probe", check_fn=check_ssrf)
 def ssrf_probe_tool(
-    endpoints_json: str,
+    endpoints: list[Endpoint],
     payloads: list[SsrfPayload] | None = None,
 ) -> list[RawFinding]:
     ...
@@ -48,10 +48,10 @@ def ssrf_probe_tool(
 
 ```python
 # wrong - the agent has no idea what strings are valid
-def ssrf_probe_tool(endpoints_json: str, payloads: list[str] | None = None) -> ...:
+def ssrf_probe_tool(endpoints: list[Endpoint], payloads: list[str] | None = None) -> ...:
 ```
 
-The `endpoints_json: str` pattern exists because CrewAI's tool-call protocol cannot pass arbitrary pydantic objects in - the wrapper takes the JSON-encoded form and unmarshals via `_parse_endpoints`. That is the only acceptable use of a bare `str` parameter for structured data; everywhere else, use the typed shape.
+The legacy `endpoints_json: str` pattern was retired in #139 once CrewAI's args-schema validation grew up enough to accept `list[<Model>]` directly. Today the LLM-side wire format is still JSON, but `args_schema.model_validate(kwargs).model_dump()` happens inside CrewAI before the function is called, so the wrapper just receives `list[dict]` and re-validates via `_parse_endpoints`. The agent sees a typed Endpoint schema instead of "pass a JSON string" - never reintroduce the JSON-string parameter for structured data.
 
 ## Workspace artifacts: writer and reader ship together
 
@@ -72,7 +72,7 @@ The reader returns the typed model; downstream agents work against the schema, n
 
 ## The `SquadMember.tools` registry
 
-Today `SquadMember.tools` is typed `list[Any]` because the local `_PentestTool` / `_ResearchBriefTool` Protocols do not share a base. That is being tightened in a follow-up (define a `CrewAITool` Protocol once, retype the registry). When that lands, the contract test that walks `MEMBER.tools` and asserts return-type annotations will be the mechanical enforcement of this skill - skill stays in AI context, test stays in CI.
+`SquadMember.tools` is typed `list[CrewAITool]` (the `CrewAITool` Protocol in `squad/__init__.py`). The local `_PentestTool` / `_ResearchBriefTool` Protocols inherit from it, so the registry is properly typed end-to-end. The contract test in `tests/test_squad_tool_contracts.py` walks every `MEMBER.tools` entry and asserts return-type annotations - skill stays in AI context, test stays in CI.
 
 ## Anti-patterns to catch
 
--- a/models/__init__.py
+++ b/models/__init__.py
@@ -165,6 +165,89 @@ class ReconResult(BaseModel):
     completed_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
 
 
+# Workspace artefacts (shared @tool wrapper return shapes)
+
+
+class RunFile(BaseModel):
+    """One file in the per-run shared workspace."""
+
+    name: str  # path relative to the run directory
+    size_bytes: int
+
+
+class RunFileContent(BaseModel):
+    """The full contents of one run-directory file."""
+
+    name: str
+    size_bytes: int
+    content: str
+
+
+# Recon-derived tool return shapes
+
+
+class OpenPortsMap(BaseModel):
+    """The recon-derived port map keyed by host.
+
+    Lives as its own model rather than ``dict[str, list[int]]`` so the
+    Penetration Tester sees a documented shape it can pattern-match on
+    when deciding which port-specific probes to run.
+    """
+
+    hosts: dict[str, list[int]] = Field(default_factory=dict)
+
+
+class LlmEndpoint(BaseModel):
+    """An endpoint flagged as LLM-backed by ``detect_llm_endpoints``.
+
+    A thin wrapper over ``Endpoint`` so the OSINT Analyst tool surface
+    carries the LLM-detection intent at the type level (the field set is
+    identical, but the model name marks the contract).
+    """
+
+    url: str
+    status_code: int | None = None
+    technologies: list[str] = Field(default_factory=list)
+    parameters: list[str] = Field(default_factory=list)
+
+
+# Triage / report / OSINT tool return shapes
+#
+# These wrap a workspace path plus the validation report produced by the
+# underlying tool. Keeping the path and the validation grouped on the same
+# model means the agent does not have to guess which keys to read on a
+# bare dict.
+
+
+class RawFindingSummary(BaseModel):
+    """Compact summary of one raw finding, returned by List Raw Findings."""
+
+    index: int
+    title: str
+    vuln_class: str
+    target: str
+    severity_hint: Severity
+    evidence_bytes: int
+
+
+class ProgrammeReportSummary(BaseModel):
+    """Compact summary of one HackerOne report listed against a programme."""
+
+    report_id: str | None = None
+    title: str | None = None
+    severity: str | None = None
+    state: str | None = None
+
+
+class CveEntry(BaseModel):
+    """One NVD CVE record, returned by NVD CVE Lookup."""
+
+    id: str
+    cvss_score: float | None = None
+    cvss_vector: str | None = None
+    description: str = ""
+
+
 # Operational metrics (emitted after every pipeline run)
 
 
--- a/squad/__init__.py
+++ b/squad/__init__.py
@@ -30,7 +30,7 @@ import functools
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, TypeVar
+from typing import Protocol, TypeVar, runtime_checkable
 
 from crewai import Agent, Task
 from crewai.tools import tool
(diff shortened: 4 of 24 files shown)


coolhandle01:
Reviewing the `cached_tool` helper specifically (the typed-return refactor that came along with it looks fine, separate concern).

I think the helper ends up duplicating something CrewAI 1.14.5 already does for free — and the new memoisation test only passes because it takes a path the agent runtime never takes.

**Where CrewAI already caches every `@tool`**

In `crewai==1.14.5` (what `pyproject.toml` pins):

- `crewai/tools/base_tool.py:80` — `_default_cache_function(_args, _result) -> bool: return True`
- `crewai/tools/base_tool.py:161` — `cache_function: SerializableCallable = Field(default=_default_cache_function, ...)`
- `crewai/agents/agent_builder/base_agent.py:234` — `cache: bool = Field(default=True, ...)`
- `crewai/agents/agent_builder/base_agent.py:686-689` — `set_cache_handler` wires `tools_handler.cache = cache_handler` when `self.cache` is True (our `build_agent` doesn't override it)

Runtime path in `crewai/tools/tool_usage.py`:

- L274-285 — before invoking the tool, `tools_handler.cache.read(tool=name, input=json.dumps(args))` runs; if it hits, `result` is set from cache and `tool.ainvoke` is skipped entirely
- L343-356 — after a successful invoke, `should_cache = cache_function(args, result)` (default True) and the result is stored under that same key

So the second time the LLM calls `lookup_cwe_tool(query="SQLi")`, CrewAI returns its cached output before `tool.func` runs at all. The `functools.cache` inside `tool.func` is never touched on that second call — CrewAI's outer cache already short-circuited it.

**Why the new test passes**

`test_lookup_cwe_tool_memoises_repeat_queries` calls `lookup_cwe_tool.func("XXE")` directly, which bypasses `tools_handler.cache` and reaches the wrapped function. The `assert first is second` check is then asserting an identity property of `functools.cache` (same list object on hit), not a behaviour the runtime relies on or even exposes — CrewAI's cache stores serialised output, so the agent-visible path doesn't preserve object identity either way.

**Two secondary nits if the helper stays**

- `functools.cache` is unbounded. Fine for the CWE/OWASP catalogues with a small query space; the docstring invites wider use, where unbounded would matter.
- It memoises object identity, so any consumer that mutates the returned list (`.pop`, `.sort`) poisons later calls. Bare `@tool` doesn't have this — `cwe_data.lookup` returns a fresh list each call and CrewAI's cache serialises before storing.

**Suggestion**

Drop `cached_tool` and go back to `@tool` on these two. If the goal is to *signal* "pure data lookup" for future readers, a marker / comment is cheaper than a wrapper and avoids the identity-sharing surprise. The behaviour the test asserts is already provided by `tools_handler.cache` at the agent-call layer, where it actually matters.

Happy to be wrong — if there's a scenario where CrewAI's outer cache is off (or the tool is invoked outside an Agent) and `functools.cache` saves real work, would be useful to call that out as the helper's reason-for-existing.

---

coolhandle01:
Pushed `fd11508` to this branch acting on this review.

**What landed**: dropped the `cached_tool` helper, restored bare `@tool` on the four lookup wrappers (`lookup_cwe_tool` / `lookup_owasp_tool` in `technical_author`, `osint_analyst`, and `vulnerability_researcher` — the latter two picked up `@cached_tool` in #139 on top of the original commit), removed the identity-asserting `test_lookup_cwe_tool_memoises_repeat_queries`, and removed the now-stale `lookup_*.func.cache_clear()` setup calls in the OSINT / VR test suites. 7 files, +14/-60.

**Why I accepted the suggestion in full**:

- The runtime-path argument is decisive. CrewAI 1.14.5's `tools_handler.cache` (read at `tool_usage.py:274-285`, written at `:343-356`) short-circuits the second `lookup_cwe_tool(query="SQLi")` call before `tool.func` runs at all, so the inner `functools.cache` saves no work on the path the agents actually take. The helper duplicated behaviour we already had for free.
- The memoisation test passed only because it called `tool.func` directly, bypassing `tools_handler.cache` and asserting an identity property (`first is second`) that CrewAI's outer cache does not preserve anyway — it stores serialised output. The test asserted a behaviour the runtime never exposes; removing it loses no signal.
- The two secondary nits (unbounded cache, object-identity sharing letting `.pop` / `.sort` poison later calls) go away with the revert too — `cwe_data.lookup` / `owasp_data.lookup` build fresh lists per call and CrewAI's cache serialises before storing.

On "is there a scenario where the outer cache is off and `functools.cache` saves real work" — I couldn't construct one that the test suite or the agent runtime actually exercises. Direct `tool.func(...)` calls in tests are the only path that would benefit, and that's testing the decorator rather than any behaviour the agents see. Not enough load-bearing to keep the helper.

The "signal pure data lookup" intent isn't carried in the diff today — happy to layer in a one-line comment on each of the four wrappers in a follow-up if that framing is worth preserving.

Six-step CI parity stack runs green locally (ruff check / format, mypy, pylint 9.87/10 unchanged from base, pytest 1064 passed at 91.5% coverage, bandit clean).

---

Issue #150, opened 2026-05-22
refactor(squad): explicit args_schema on triage, report, and workspace tools


coolhandle01:
## Motivation

Final pass to complete the `args_schema` alignment kicked off by #143. The Vulnerability Researcher, Technical Author, and shared workspace tools are internal-effect (they don't fire requests at live programmes), but a mis-call here corrupts the artefacts that downstream stages depend on: a misread CVSS vector flips a report's severity; a wrong `relative_path` on `Read Run File` returns the wrong workspace blob and the subsequent agent reasons over the wrong inputs.

The cost of a mis-call is less visible than the PT / OSINT / H1-API surfaces but it is structurally identical: the schema is the contract the LLM is shown when picking the tool, and the explicit path attaches per-field descriptions the inferred path cannot.

## What changes

Every `@tool`-decorated function in the following modules gets an explicit Pydantic `args_schema`:

- **VR (`squad/vulnerability_researcher/__init__.py`):** `NVD CVE Lookup`, `List Programme Reports`, `Finalise Research`, `List Raw Findings`, `Read Raw Finding`, `Calculate CVSS Score`, `Lookup CWE`, `Lookup OWASP Guidance`, `Assess Raw Finding`, `Discard Finding`, `Finalise Triage`
- **TA (`squad/technical_author/__init__.py`):** `Sanitise Evidence`, `Lookup CWE`, `Lookup OWASP Guidance`, `Calculate CVSS Score`, `List Programme Reports`, `Draft Vulnerability Report`, `Finalise Reports`
- **Workspace (`squad/workspace_tools.py`):** `List Run Files`, `Read Run File`, `Read Attack Plan`

The shared lookup tools (`Lookup CWE`, `Lookup OWASP Guidance`, `Calculate CVSS Score`) are declared per-agent and currently have bespoke docstrings; this issue does **not** consolidate them - that is a separate refactor. Each agent's copy gets its own `_<AgentInitials><ToolName>Args` class so the contract test can keep the registry-and-mapping check from `tests/test_pentest_args_schemas.py`.

For `Calculate CVSS Score` the schema is where the CVSS vector format (`CVSS:3.1/AV:N/AC:L/...`) gets documented. The inferred docstring path made this hard to see; this is the field-description sweet spot.

## Constraints

- **No behaviour changes.** Triage, report drafting, and workspace IO outputs are identical.
- **Bare `@tool` stays.** These wrappers are not probe-style.
[...]

Issue #147, opened 2026-05-22
refactor(pt-cloud): explicit args_schema on cloud / infra check tools


coolhandle01:
## Motivation

Direct sibling of #143. The cloud / infra `@tool` wrappers on the Penetration Tester (`s3_check_tool`, `mongodb_tool`, `cpanel_tool`, ...) were intentionally left out of #143 because the issue scoped specifically to `@pentest_tool` probes. The framing #143 used - "probes hit live services on live programmes, mis-targeting costs money, generates noise the operator has to apologise for, and at the limit risks the squad's relationship with the H1 programme" - applies identically to these tools: they all fire HTTP / TCP probes at hosts the OSINT Analyst surfaced. A mis-call costs the same.

The inferred path crams every parameter into the docstring and gives no place for per-field descriptions; the explicit path can address fields individually.

## What changes

Every `@tool`-decorated wrapper in `squad/penetration_tester/__init__.py` that is **not** a `@pentest_tool` probe and **not** in the `out` set below gets an explicit Pydantic `args_schema`:

- `s3_check_tool`, `azure_storage_check_tool`
- `elasticsearch_tool`, `couchdb_tool`, `redis_tool`, `mongodb_tool`, `postgresql_tool`, `mysql_tool`
- `sensitive_files_tool`, `admin_panels_tool`
- `cpanel_tool`, `plesk_tool`, `directadmin_tool`, `webmin_tool`
- `grafana_tool`, `kibana_tool`, `portainer_tool`, `consul_vault_tool`

The parameter shape is narrow: every cloud / infra tool takes either `recon_path: str` or `endpoints: list[Endpoint]`, so the schema classes are small. The reason for the per-tool description (rather than a shared `_ReconPathArgs`) is that the description is what the LLM reads when picking the tool - which host class to focus on, when this tool is worth firing, whether the port should already be open per recon. Tailored is the point.

## Constraints

- **No behaviour changes.** Wrapper bodies stay the same.
- **Bare `@tool` stays.** These wrappers are not probe-style; they do not need `@pentest_tool` and the StrEnum ceremony. They just need an explicit `args_schema`.
- **Skill update required.** `cybersquad-tool` (the universal skill) gains an `args_schema` note - the requirement is no longer specific to pentest probes once this lands. `cybersquad-pentest-tool` already documents it.
[...]

Issue #144, opened 2026-05-22
feat(MCP) - fleshing out crew AI implementation


coolhandle01:
## More Skills!

I'd like to try these popular skills to see if they aid in developing cybersquad:
https://mcpmarket.com/tools/skills/python-testing-and-pytest-patterns
https://mcpmarket.com/tools/skills/python-patterns-best-practices-7
https://mcpmarket.com/tools/skills/postgresql-patterns-1777849920215 (`PGSearchTool`)

## MCPs for cybersquad!

I think these are worth investigating with a view to fleshing out cybersquad agent capabilities:
- [x] https://github.com/modelcontextprotocol/servers/tree/main/src/time
- [x] https://mcpmarket.com/server/burp-suite
- [x] https://mcpmarket.com/server/playwright-2
- [x] https://mcpmarket.com/server/cloudflare-playwright
- [ ] https://github.com/swimmwatch/cloakbrowser-mcp

coolhandle01:
Regrouping by the change boundary each idea actually lives on (runtime capability vs pipeline plumbing vs contributor tooling) rather than by docs section. Some items belong in more than one bucket; I've put them where the work would land.

## 1. Runtime agent capabilities (tools the squad uses)

### CrewAI built-in tools

Worth pulling in:

- **`JSONSearchTool`** - good fit. The handoff artefacts (`select.json`, `recon.json`, `verified.json`, etc.) already cross agent boundaries via the workspace pattern. JSON search lets downstream agents query upstream output without re-reading the whole file into the prompt. Best win: research and pentest querying recon.
- **`SerperDevTool`** - useful for the OSINT Analyst, but adds a paid API dependency to the critical path. Should be gated behind an env var + capability check (skip if `SERPER_API_KEY` is unset) the same way `h1_api` is gated.
- **`ScrapeWebsiteTool`** - overlaps heavily with `tools/recon/` (`http.py`, `html.py`, `waybackurls.py`). Only worth adding if we want CrewAI's auto-paginated scrape semantics; otherwise redundant.

Worth avoiding:

- **`FileReadTool` / `FileWriteTool`** - these undermine the writer/reader workspace pair pattern the `cybersquad-tool` contributor skill enforces. The current pattern keeps workspace boundaries auditable (member X writes its own workspace only). Generic file read/write would collapse that. Veto unless we deliberately change the pattern.

### Tool caching

`cache_function` on `@tool` is appealing but context-sensitive:

[...]

PR #152, opened 2026-05-24
refactor(osint): explicit args_schema on every OSINT recon tool (#148)

### LINKED ISSUE #148: refactor(osint): explicit args_schema on every OSINT recon tool
## Motivation

Follow-on from #143. OSINT recon tools fire `subfinder`, `httpx`, `nmap`, `testssl.sh`, `dirfuzz`, `waybackurls`, cert-transparency lookups, tracepath, and LLM endpoint detection against in-scope hosts. The same "live programmes / mis-call costs money" argument that motivated explicit `args_schema` on the PT probes (see #143) applies here, plus a scope-leak risk that is specific to recon: a mis-targeted `host_filter` or wrong `host_contains` substring can pull a sibling programme's endpoints into the run and silently corrupt every downstream stage.

The inferred schema cannot attach per-field descriptions; the explicit path can. The OSINT tools rely especially on filter-parameter discipline - `host_filter`, `status`, `tech`, `host_contains` etc. - and those filters are exactly the fields where targeted guidance pays off.

## What changes

Every `@tool`-decorated function in `squad/osint_analyst/__init__.py` gets an explicit Pydantic `args_schema`:

- `Run Initial Sweep` (the umbrella sweep)
- `Recon Subdomains`, `Recon Endpoints`, `Recon Open Ports`
- `Certificate Transparency Lookup`
- `Historical URL Discovery`
- `LLM Endpoint Detection`
- `Probe Hostnames`
- `Detect Takeover Candidates`
- `Lookup CWE`, `Lookup OWASP Guidance` (OSINT's copies)
- `Annotate Host`
- `Uncovered Hosts`
- `Finalise Recon`

Schema classes live inline above each wrapper. Where filter parameters narrow the surface (`host_filter`, `host_contains`, `status`, `tech`, `offset`, `limit`), the `Field(description=...)` documents:

- what the filter matches (substring vs. exact, case sensitivity)
- when to pass it vs. omit
- the request-cost / scope-leak implication of leaving it broad

## Constraints

- **No behaviour changes.** Probe execution and recon outputs are identical.
- **Bare `@tool` stays.** OSINT tools are not probe-style; they don't need `@pentest_tool`.
- **Skill update.** `cybersquad-tool` (universal) gains the `args_schema` note if not already landed by a sibling follow-up.
- **Tests follow.** Per-tool schema accept/reject tests, structurally aligned with `tests/test_pentest_args_schemas.py`. The `Lookup CWE` / `Lookup OWASP Guidance` schemas may be shared with the VR / TA copies of those tools (out of scope for this issue, see the triage/report follow-up).

## Scope

In: every `@tool` in `squad/osint_analyst/__init__.py`.

Out: tools defined in `squad/workspace_tools.py` (covered separately).

## Dependencies / ordering

Independent of the PT-cloud and H1-API follow-ups. The first OSINT / PT-cloud / H1-API PR to land owns the `cybersquad-tool` skill edit; the others inherit it.

## Acceptance criteria

- [ ] Every OSINT `@tool` has an explicit `args_schema` Pydantic model.
- [ ] Every schema field carries a `Field(description=...)` written as agent-facing targeting guidance, with filter semantics (substring vs. exact, case sensitivity) explicit.
- [ ] `cybersquad-tool` skill updated to make `args_schema` part of the universal `@tool` pattern (or note already covered if a sibling issue landed it first).
- [ ] Per-tool contract tests for schema accept/reject.
- [ ] Coverage floor (`fail_under=90`) holds.
- [ ] No behaviour changes; recon outputs unchanged.

coolhandle01:
## Summary

Closes #148. Extends the args_schema discipline from #143 / #146 / #147 to the 14 OSINT Analyst `@tool` wrappers covering subfinder / httpx / nmap / ffuf / crt.sh / waybackurls / dnsx / LLM endpoint detection / recon-query slicers / host annotation / finalisation.

**Mea culpa on scope**: the PR title is "OSINT args_schema sweep" but the diff is larger than that suggests. Every addition is review-driven (the comment thread is the audit trail) and conceptually tied to the args_schema work, but I want to be honest about how we got from "sweep 14 wrappers" to ~1300 line diff. See the timeline at the end if you want the gory details.

### OSINT args_schema sweep (the original #148 scope)

14 OSINT wrappers swept to `@cyber_tool` with inline `_<Name>Args` schemas. Field descriptions written as agent-facing targeting guidance, with particular care on recon filter parameters (`host_filter`, `host_contains`, `status`, `tech`) where mis-targeting is the scope-leak risk specific to recon.

The 2 remaining `@tool` wrappers on OSINT (workspace readers re-exported from `squad.workspace_tools`) stay inferred-path - covered by #150.

### Helper rename: `typed_tool` -> `cyber_tool`

Picked up mid-sweep on review feedback. `cyber_tool` makes the brand visible at every consumer site; the helper is conceptually "the cybersquad-blessed `@tool` replacement". sed across 6 files (definition, PT module, OSINT module, the cybersquad-tool skill, both args_schema test files). No behaviour change.

### `endpoints_json: str` -> `list[Endpoint]` on `llm_detection_tool`

The last surviving instance of the #139-retired stringified-JSON pattern. Migrated to the canonical `list[Endpoint]` shape with the both-shape adapter pattern (`Endpoint.model_validate(e)` accepts both Endpoint instances and dicts; covers runtime + direct-test paths).

### `Hostname` typed-string primitive

New `Hostname = Annotated[str, AfterValidator(_validate_hostname)]` in `models/primitives.py`. RFC 1123 hostname validation - rejects schemes, ports, paths, empty input, leading/trailing-hyphen labels, oversized labels (>63 chars), oversized total (>253 chars), garbage chars. Lowercases + strips on the way through.

Applied across 6 fields on 5 OSINT schemas (Probe Hostnames, Detect Takeover Candidates, Recon Open Ports, Annotate Host, Cert Transparency, Historical URL Discovery) plus composed into `HostInsight.hostname`, `OpenPortsMap.hosts` keys, `ReconResult.subdomains` / `open_ports` keys / `network_hops` keys.

### `HttpUrl` typed-string primitive

Light flavour: `Annotated[str, AfterValidator(_validate_endpoint_url)]`. Parses with `urlparse`, asserts http/https scheme, runs the embedded hostname through the `Hostname` validator. Runtime type stays `str` so existing `endpoint.url.startswith(...)` call sites keep working.

Applied to `Endpoint.url` and `LlmEndpoint.url`. FIXME comments in `primitives.py` and `asset.py` point at Pydantic's built-in `HttpUrl` as the stronger future contract, gated on auditing every consumer.

### StrEnum on Annotate Host

`_AnnotateHostArgs.role` / `.priority` typed as `HostRole` / `HostPriority` StrEnums directly (was `str`). Schema rejects unknown values upstream of the wrapper's coercion. Signature updated to match.

### Models package split

`models/__init__.py` reduced from 319 lines to a pure re-export surface (all `from models import X` consumer imports unchanged). Per-domain modules:

- `models/primitives.py` - Severity, Hostname, HttpUrl
- `models/finding.py` - RawFinding, VerifiedVulnerability, RawFindingSummary
- `models/asset.py` - Endpoint, EndpointPage, HostRole, HostPriority, HostInsight, OpenPortsMap, LlmEndpoint, ReconResult
- `models/workspace.py` - RunFile, RunFileContent
- `models/cve.py` - CveEntry
- `models/metrics.py` - RunMetrics
- ProgrammeReportSummary moved to `models/h1.py` (it is H1-shaped)

Pre-split circular-import dance (h1 importing Severity while __init__ initialised) removed; dependency layers flow primitives -> finding -> h1 -> asset.

### LlmEndpoint FIXME

Class kept as-is for now but carries a FIXME pointing at #144 (MCP discipline) and #83 (framework probes). The realisation surfaced during review: `LlmEndpoint` today is `Endpoint` with a docstring change, and what the original author was reaching for is properly the `DiscoveredMCP` shape #144 sketched. The class will migrate into `models/framework.py` per #83 as the first framework-typed model. Until then, do not extend it as a generic LLM endpoint shape.

### Skill refresh

Both contributor skills brought back into alignment with the codebase:

- **`cybersquad-tool`**: new "Typed string primitives" section (Hostname + HttpUrl), new "Where models live" section with the per-domain layout table, all example code blocks switched from bare `@tool` to `@cyber_tool` / `@pentest_tool`, anti-patterns gained "bare str on hostname / URL fields", "dict[str, ...] keys that should be Hostname", "adding to __init__.py". Canonical examples updated (OSINT joins PT as fully migrated; `read_attack_plan_tool` is the typed-reader example, not the backlog target).
- **`cybersquad-pentest-tool`**: two rename fixes (`test_pentest_args_schemas.py` -> `test_pt_args_schemas.py`, `_PROBE_SCHEMAS` -> `_PT_SCHEMAS`).

### Contract tests

- `tests/test_osint_args_schemas.py` mirrors `test_pt_args_schemas.py`: closed-world `_OSINT_SCHEMAS` mapping, every-field-has-description check, parametrized accept / reject cases for the URL-less schemas, fixture-using dedicated methods for the URL / hostname-shaped ones.
- `tests/test_models.py` `TestHostname` and `TestHttpUrl` exhaustively cover both validators (accept paths from the `victim_url` fixture, reject corpus derived from the same fixture so test intent reads as "in-scope target, deliberately mis-shaped").
- `tests/test_recon_insights.py` two tests updated for the upstream-validation change (HostInsight construction now rejects empty / mis-shaped hostnames; the `validate_insight` branch that handled them is now dead defensive code).

## Out of scope (filed as follow-ups)

- **#153** - `@cyber_tool(scope_filter=...)` check_fn for hostname-taking tools (the "scope_guard as check_fn" Frankensteiney pattern is real but contained; lifted out into its own focused PR).
- Broader OSINT API refactor (programme_handle threading, sweep_path implicit, HostInsight-shaped Annotate Host input, deduplicated CWE / OWASP lookups across agents) - user is taking that on directly.
- `LlmEndpoint` -> `DiscoveredMCP` migration into `models/framework.py` - flagged via FIXME, sequenced after #83 lands the framework-probe infrastructure.
- Migration of `Endpoint.url` from the light validator to Pydantic's built-in `HttpUrl` - FIXME comment + audit of every consumer-side `url.startswith(...)` site needed first.

## Test plan

- [x] `ruff check .` - clean
- [x] `pylint .` - 9.87/10 (no regression)
- [x] `pytest -m unit --cov` - 1320 passed, 91.73% coverage
- [x] OSINT registry post-sweep: 14 explicit schemas, 2 inferred (workspace readers, out of scope)

## Commit timeline (for the curious)

The review thread tells the full story; the short version of how scope accreted:

1. `56ef905` - the actual #148 sweep + `typed_tool` rename to `cyber_tool`.
2. `9da0647` - `endpoints_json: str` migration on `llm_detection_tool` (you flagged the anti-pattern in flight).
3. `cb289c3` - StrEnum on `role` / `priority` + fixture-using test for LlmDetection (small review fixes).
4. `9406fa9` - URL fixtures replacing example.com literals across tests (review clarification).
5. `5a0eac1` - `Hostname` typed-string primitive + composition through OSINT schemas (you talked me into it).
6. `c06143f` - Hostname tests derived from victim_url fixture (review fix).
7. `9e21166` - models package split + `HttpUrl` primitive + `Hostname` composition through `HostInsight` / `OpenPortsMap` / `ReconResult` (the big one).
8. `6ac16bb` - cybersquad-tool + cybersquad-pentest-tool skill refresh (audit caught both stale).
9. `d0bd5c3` - LlmEndpoint FIXME pointing at #144 / #83 / `models/framework.py` (review surfaced the design lineage).

Each step was a small, defensible expansion in response to specific review feedback. The cumulative diff is bigger than "OSINT args_schema sweep" implies, but the connective tissue is real: every addition tightens or composes the typed-input discipline this PR series exists to land.

### DIFF
--- a/squad/osint_analyst/__init__.py
+++ b/squad/osint_analyst/__init__.py
@@ -9,6 +9,7 @@ from models import (
     Endpoint,
     EndpointPage,
     HostInsight,
+    Hostname,
     HostPriority,
     HostRole,
     LlmEndpoint,
@@ -204,18 +205,21 @@ class _ReconOpenPortsArgs(BaseModel):
         default="sweep.json",
         description="Relative path to the OA's sweep file. Defaults to ``sweep.json``.",
     )
-    host: str | None = Field(
+    host: Hostname | None = Field(
         default=None,
         description=(
             "Exact hostname to restrict the result to (no wildcards, no"
-            " substring). Omit or pass null to get the full per-host port"
-            " map. Use to look up a single host before annotating."
+            " substring). Validated as an RFC 1123 hostname - URLs / ports"
+            " / paths reject upstream. Omit or pass null to get the full"
+            " per-host port map."
         ),
     )
 
 
 @cyber_tool("Recon Open Ports", args_schema=_ReconOpenPortsArgs)
-def recon_open_ports_tool(sweep_path: str = "sweep.json", host: str | None = None) -> OpenPortsMap:
+def recon_open_ports_tool(
+    sweep_path: str = "sweep.json", host: Hostname | None = None
+) -> OpenPortsMap:
     """
     Return the open-port map per host from the sweep. Passing a ``host``
     restricts the result to that single host. Use to surface non-HTTP
@@ -228,11 +232,11 @@ def recon_open_ports_tool(sweep_path: str = "sweep.json", host: str | None = Non
 class _CertTransparencyArgs(BaseModel):
     """Explicit args_schema for the Certificate Transparency Lookup tool (#148)."""
 
-    domain: str = Field(
+    domain: Hostname = Field(
         description=(
             "Apex domain to look up in crt.sh certificate transparency logs"
-            " (e.g. 'example.com'). Returns historical subdomains across all"
-            " certificates issued for the domain. Pass an apex, not a"
+            " (e.g. 'example.com'). Validated as an RFC 1123 hostname"
+            " (URLs / ports / paths reject upstream). Pass an apex, not a"
             " subdomain - crt.sh returns the broader set when queried on"
             " the apex."
         ),
@@ -240,7 +244,7 @@ class _CertTransparencyArgs(BaseModel):
 
 
 @cyber_tool("Certificate Transparency Lookup", args_schema=_CertTransparencyArgs)
-def cert_transparency_tool(domain: str) -> list[str]:
+def cert_transparency_tool(domain: Hostname) -> list[str]:
     """
     Query crt.sh certificate transparency logs to discover subdomains not
     found by active enumeration. Returns deduplicated hostnames. Feed
@@ -252,18 +256,19 @@ def cert_transparency_tool(domain: str) -> list[str]:
 class _HistoricalUrlsArgs(BaseModel):
     """Explicit args_schema for the Historical URL Discovery tool (#148)."""
 
-    domain: str = Field(
+    domain: Hostname = Field(
         description=(
-            "Domain to query waybackurls for (apex or subdomain). Returns"
-            " historical URLs the Wayback Machine has archived. Many paths"
-            " will be 404s today; feed candidates to Probe Hostnames to"
-            " confirm liveness before annotating."
+            "Domain to query waybackurls for (apex or subdomain). Validated"
+            " as an RFC 1123 hostname (URLs / ports / paths reject"
+            " upstream). Returns historical URLs the Wayback Machine has"
+            " archived. Many paths will be 404s today; feed candidates to"
+            " Probe Hostnames to confirm liveness before annotating."
         ),
     )
 
 
 @cyber_tool("Historical URL Discovery", args_schema=_HistoricalUrlsArgs)
-def historical_urls_tool(domain: str) -> list[str]:
+def historical_urls_tool(domain: Hostname) -> list[str]:
     """
     Use waybackurls to find historical endpoints for a domain from the
     Wayback Machine. Surfaces paths that may no longer be linked but still
@@ -305,15 +310,17 @@ def llm_detection_tool(endpoints: list[Endpoint]) -> list[LlmEndpoint]:
 class _ProbeHostnamesArgs(BaseModel):
     """Explicit args_schema for the Probe Hostnames tool (#148)."""
 
-    hostnames: list[str] = Field(
+    hostnames: list[Hostname] = Field(
         description=(
             "Hostnames to re-probe with httpx for liveness, status code,"
-            " and technology fingerprinting. Typically the net-new ones"
-            " surfaced by Certificate Transparency or Historical URL"
-            " Discovery that were missed by the initial sweep. Each"
-            " hostname is scope-filtered against the programme before"
-            " any HTTP traffic fires - out-of-scope hostnames are"
-            " dropped silently."
+            " and technology fingerprinting. Each entry is validated as an"
+            " RFC 1123 hostname; URLs / ports / paths reject upstream"
+            " (catches the common 'agent handed us a URL when we asked for"
+            " a hostname' case before the scope filter silently drops it)."
+            " Typically the net-new ones surfaced by Certificate"
+            " Transparency or Historical URL Discovery that were missed by"
+            " the initial sweep. After this, each hostname is scope-"
+            "filtered against the programme before any HTTP traffic fires."
         ),
     )
     programme_handle: str = Field(
@@ -327,7 +334,7 @@ class _ProbeHostnamesArgs(BaseModel):
 
 
 @cyber_tool("Probe Hostnames", args_schema=_ProbeHostnamesArgs)
-def probe_hostnames_tool(hostnames: list[str], programme_handle: str) -> list[Endpoint]:
+def probe_hostnames_tool(hostnames: list[Hostname], programme_handle: str) -> list[Endpoint]:
     """
     Re-probe a list of hostnames with httpx to confirm liveness, capture
     status codes, and fingerprint technologies. Use this on hostnames
@@ -352,13 +359,15 @@ def probe_hostnames_tool(hostnames: list[str], programme_handle: str) -> list[En
 class _DetectTakeoverCandidatesArgs(BaseModel):
     """Explicit args_schema for the Detect Takeover Candidates tool (#148)."""
 
-    hostnames: list[str] = Field(
+    hostnames: list[Hostname] = Field(
         description=(
             "Hostnames to resolve via dnsx and flag for subdomain takeover."
-            " A candidate fires when the CNAME points to a known-vulnerable"
-            " provider (AWS S3, Heroku, GitHub Pages, Azure, Vercel,"
-            " Netlify, ...) or when the CNAME chain dangles. Hostnames are"
-            " scope-filtered before any DNS traffic fires."
+            " Each entry is validated as an RFC 1123 hostname; URLs / ports"
+            " / paths reject upstream. A candidate fires when the CNAME"
+            " points to a known-vulnerable provider (AWS S3, Heroku,"
+            " GitHub Pages, Azure, Vercel, Netlify, ...) or when the CNAME"
+            " chain dangles. Hostnames are scope-filtered before any DNS"
+            " traffic fires."
         ),
     )
     programme_handle: str = Field(
@@ -372,7 +381,7 @@ class _DetectTakeoverCandidatesArgs(BaseModel):
 
 @cyber_tool("Detect Takeover Candidates", args_schema=_DetectTakeoverCandidatesArgs)
 def detect_takeover_candidates_tool(
-    hostnames: list[str], programme_handle: str
+    hostnames: list[Hostname], programme_handle: str
 ) -> list[TakeoverCandidate]:
     """
     Resolve each hostname via dnsx and flag subdomain-takeover candidates.
@@ -452,11 +461,12 @@ def lookup_owasp_tool(query: str) -> list[OWASPEntry]:
 class _AnnotateHostArgs(BaseModel):
     """Explicit args_schema for the Annotate Host tool (#148)."""
 
-    hostname: str = Field(
+    hostname: Hostname = Field(
         description=(
             "Hostname to annotate. Must already be in the sweep, or have"
             " been surfaced by Certificate Transparency / Historical URL"
-            " Discovery / Probe Hostnames. Lowercased before save."
+            " Discovery / Probe Hostnames. Validated as an RFC 1123"
+            " hostname (URLs / ports / paths reject upstream)."
         ),
     )
     role: HostRole = Field(
@@ -509,7 +519,7 @@ class _AnnotateHostArgs(BaseModel):
 
 @cyber_tool("Annotate Host", args_schema=_AnnotateHostArgs)
 def annotate_host_tool(
-    hostname: str,
+    hostname: Hostname,
     role: HostRole,
     priority: HostPriority,
     notes: str,
--- a/squad/osint_analyst/__init__.py
+++ b/squad/osint_analyst/__init__.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import json
 from pathlib import Path
 
-from crewai.tools import tool
+from pydantic import BaseModel, Field
 
 import runtime
 from models import (
@@ -18,7 +18,7 @@ from models import (
     OpenPortsMap,
 )
 from models.h1 import Programme
-from squad import SquadMember, read_run_file_tool, read_run_filelist_tool
+from squad import SquadMember, cyber_tool, read_run_file_tool, read_run_filelist_tool
 from tools import http
 from tools.cwe_data import CWEEntry
 from tools.cwe_data import lookup as cwe_lookup
@@ -52,7 +52,21 @@ from tools.recon_insights import (
 )
 
 
-@tool("Run Initial Sweep")
+class _RunInitialSweepArgs(BaseModel):
+    """Explicit args_schema for the Run Initial Sweep tool (#148)."""
+
+    programme_handle: str = Field(
+        description=(
+            "Exact HackerOne programme handle as it appears in the URL"
+            " (lowercase, no slashes, no spaces). The sweep runs subfinder,"
+            " httpx, nmap, ffuf, and passive TLS / DNS checks against the"
+            " programme's structured scope - the handle is the authoritative"
+            " key the H1 API uses to look the scope up."
+        ),
+    )
+
+
+@cyber_tool("Run Initial Sweep", args_schema=_RunInitialSweepArgs)
 def run_initial_sweep_tool(programme_handle: str) -> str:
     """
     Run the initial OSINT sweep against the in-scope assets of the given
@@ -77,7 +91,30 @@ def run_initial_sweep_tool(programme_handle: str) -> str:
     return "sweep.json"
 
 
-@tool("Recon Subdomains")
+class _ReconSubdomainsArgs(BaseModel):
+    """Explicit args_schema for the OSINT Recon Subdomains tool (#148)."""
+
+    sweep_path: str = Field(
+        default="sweep.json",
+        description=(
+            "Relative path to the OA's in-progress sweep file. Defaults to"
+            " ``sweep.json`` (the canonical name written by Run Initial"
+            " Sweep). Override only when re-inspecting a sweep written"
+            " under a different name."
+        ),
+    )
+    host_filter: str | None = Field(
+        default=None,
+        description=(
+            "Case-insensitive substring match against each subdomain"
+            " (e.g. 'api' returns every subdomain containing 'api',"
+            " regardless of position). Omit or pass null to get the full"
+            " list. Useful for scoping a per-role annotation pass."
+        ),
+    )
+
+
+@cyber_tool("Recon Subdomains", args_schema=_ReconSubdomainsArgs)
 def recon_subdomains_tool(
     sweep_path: str = "sweep.json", host_filter: str | None = None
 ) -> list[str]:
@@ -90,7 +127,54 @@ def recon_subdomains_tool(
     return recon_subdomains(sweep_path, host_filter=host_filter)
 
 
-@tool("Recon Endpoints")
+class _ReconEndpointsArgs(BaseModel):
+    """Explicit args_schema for the OSINT Recon Endpoints tool (#148)."""
+
+    sweep_path: str = Field(
+        default="sweep.json",
+        description="Relative path to the OA's sweep file. Defaults to ``sweep.json``.",
+    )
+    status: int | None = Field(
+        default=None,
+        description=(
+            "Exact HTTP status code to filter by (e.g. 200 for live, 401 /"
+            " 403 for auth-shaped, 301 / 302 for redirectors). Omit or pass"
+            " null for any status."
+        ),
+    )
+    tech: str | None = Field(
+        default=None,
+        description=(
+            "Case-insensitive substring match against each endpoint's"
+            " technologies list (e.g. 'wordpress' matches 'WordPress 6.4')."
+            " Use when scoping a per-stack annotation pass."
+        ),
+    )
+    host_contains: str | None = Field(
+        default=None,
+        description=(
+            "Case-insensitive substring match against the endpoint URL. Use"
+            " to narrow to a hostname or path family."
+        ),
+    )
+    offset: int = Field(
+        default=0,
+        description=(
+            "Zero-based row offset for paging. Use with ``limit`` to walk large result sets."
+        ),
+    )
+    limit: int = Field(
+        default=50,
+        description=(
+            "Maximum number of endpoints to return in this page. The OA's"
+            " context budget rewards small pages; default 50 is usually"
+            " right - go larger only when the conjunctive filters above"
+            " already narrowed the result heavily."
+        ),
+    )
+
+
+@cyber_tool("Recon Endpoints", args_schema=_ReconEndpointsArgs)
 def recon_endpoints_tool(
     sweep_path: str = "sweep.json",
     status: int | None = None,
@@ -116,7 +200,24 @@ def recon_endpoints_tool(
     )
 
 
-@tool("Recon Open Ports")
+class _ReconOpenPortsArgs(BaseModel):
+    """Explicit args_schema for the OSINT Recon Open Ports tool (#148)."""
+
+    sweep_path: str = Field(
+        default="sweep.json",
+        description="Relative path to the OA's sweep file. Defaults to ``sweep.json``.",
+    )
+    host: str | None = Field(
+        default=None,
+        description=(
+            "Exact hostname to restrict the result to (no wildcards, no"
+            " substring). Omit or pass null to get the full per-host port"
+            " map. Use to look up a single host before annotating."
+        ),
+    )
+
+
+@cyber_tool("Recon Open Ports", args_schema=_ReconOpenPortsArgs)
 def recon_open_ports_tool(sweep_path: str = "sweep.json", host: str | None = None) -> OpenPortsMap:
     """
     Return the open-port map per host from the sweep. Passing a ``host``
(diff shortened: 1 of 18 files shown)


squad/osint_analyst/__init__.py
```diff
@@ -236,7 +449,63 @@ def lookup_owasp_tool(query: str) -> list[OWASPEntry]:
...
+        ),
+    )
+    role: str = Field(
+        description=(
+            "One of: admin, api, auth, app, cdn, static, mail, infra, dev,"
+            " unknown. Drives downstream prioritisation - admin / auth"
+            " hosts attract more probe budget than static / cdn ones."
+        ),
+    )
+    priority: str = Field(
```
coolhandle01: This `priority` should probably be an enum (`StrEnum`?)

squad/osint_analyst/__init__.py
```diff
@@ -236,7 +449,63 @@ def lookup_owasp_tool(query: str) -> list[OWASPEntry]:
...
+            " Boot'). The quality gate warns when the sweep saw tech the"
+            " annotation drops - the annotation is meant to be a curation"
+            " of what was seen, not an alternate inventory."
+        ),
+    )
+    sweep_path: str = Field(
+        default="sweep.json",
+        description="Relative path to the OA's sweep file. Defaults to ``sweep.json``.",
+    )
+    programme_handle: str | None = Field(
```
coolhandle01, line 510: I mean, this would be the same on every entry no? It's recon of one programme's scope.

If we need this, it should be coming from programme.json?

I did think it was weird that one of the tools had a Field/input for programme.

The scope guard might be a bit wonky. Ideally, it's a function that provided some list[Endpoints] returns fase for or filters out Endpoints that aren't in programme.json's scope definition. Maybe I misunderstood but I was thinking that the scope_guard was going to be something that could be used for a check_fn on @tools (originally with @pentest_tools) where there was a list[Endpoint] input 

I think the OSINTs tools may be overly simplified, too. I fear now may be the time for a serious refactor of the OSINT Analyst whilst the rules are allowed to change - now that the pattern is known for "how to make a tool the LLM can really understand", maybe the current OA setup is now looking a bit naive?

tests/test_osint_args_schemas.py
```diff
@@ -0,0 +1,212 @@
...
+            (_ReconSubdomainsArgs, {"sweep_path": "sweep.json", "host_filter": "api"}),
+            (_ReconEndpointsArgs, {}),
+            (
+                _ReconEndpointsArgs,
+                {"status": 200, "tech": "wordpress", "host_contains": "admin", "limit": 25},
+            ),
+            (_ReconOpenPortsArgs, {}),
+            (_ReconOpenPortsArgs, {"host": "victim.example.com"}),
+            (_CertTransparencyArgs, {"domain": "example.com"}),
+            (_HistoricalUrlsArgs, {"domain": "example.com"}),
```
coolhandle01: Should use fixtures for endpoints

tests/test_squad_osint_analyst.py
```diff
@@ -349,12 +348,15 @@ def test_llm_detection_tool(self, endpoint) -> None:
         from models import LlmEndpoint
         from squad.osint_analyst import llm_detection_tool
 
-        endpoints_json = json.dumps([endpoint.model_dump(mode="json")])
+        # CrewAI's args_schema validation hands us list[dict] at runtime;
```
coolhandle01, line 351: > validation hands us list[dict] at runtime

.. what we're doing here, it is *definitely* sane and valid, right? We're definitely not over-engineering with this refactor?

squad/osint_analyst/__init__.py
coolhandle01: Done in `cb289c3` - both `role` and `priority` now typed as `HostRole` / `HostPriority` directly on `_AnnotateHostArgs` and the function signature. Agent sees the enum values in the generated JSON schema; unknown values reject at schema validation time instead of at the `HostRole(role)` coercion in the body. Idempotent `HostRole(role)` / `HostPriority(priority)` calls in the body kept (they handle either string or enum input, resilient to whichever shape CrewAI's serialisation hands the wrapper).

Added a contract test (`test_annotate_host_rejects_unknown_role_and_priority`) documenting the upstream-validation behaviour.

---

tests/test_osint_args_schemas.py
coolhandle01: Done in `cb289c3` - dropped the empty-list parametrize entry for `_LlmDetectionArgs` and added `test_llm_detection_accepts_populated_endpoint_list` which uses the `endpoint` fixture from `conftest.py`. The fixture-backed test exercises the real `Endpoint` shape (URL, status_code, technologies, parameters) through `model_validate`, so the schema's field-level type validation is now part of the contract test rather than just "accepts list".

Worth noting: the parametrize cases that pass `{"endpoints": []}` on `_SsrfArgs`, `_SqlmapArgs` etc. in `test_pt_args_schemas.py` have the same gap - they prove the field accepts a list but don't exercise a populated Endpoint. Out of scope to chase them here (#147 is merged), but worth flagging for the cleanup pass we discussed.

---

tests/test_squad_osint_analyst.py
coolhandle01: Honest assessment: yes, sane and not over-engineering - this is the canonical pattern from PT, documented in `_parse_endpoints` in `squad/penetration_tester/__init__.py` and lifted verbatim here.

The mechanism is real:

1. **Runtime path**: CrewAI receives the LLM's tool call as JSON, runs `args_schema.model_validate(...).model_dump()` to validate the shape, then hands the wrapper `list[dict]`. The pydantic instances are gone by the time our body executes.
2. **Test path**: direct invocations call `tool.func([endpoint])` and pass typed `Endpoint` instances straight through, since we're not going through CrewAI's serialisation layer.
3. **`Endpoint.model_validate(e)` accepts both**: it returns the same `Endpoint` when given an instance, constructs a new one when given a dict. So one helper covers both call paths with no branching.

The alternative would be either (a) accept only `Endpoint` instances and break the runtime path (LLM calls would fail because they hand us dicts), or (b) accept only dicts and break the direct test invocations (tests would need to round-trip through `model_dump`). The two-line adapter avoids both.

The comment in the test is genuinely about the *runtime* shape - that's what `list[dict]` refers to. The behavioural test exercises the dict path because that's the higher-risk one (it's what the LLM actually does); the args_schema contract test exercises the populated-Endpoint path separately.

So: not over-engineering, just paying the cost of CrewAI's serialisation boundary. The doc would feel more excessive if we had a custom adapter; using `model_validate` (which exists for exactly this both-shape case) keeps it cheap.

---

squad/osint_analyst/__init__.py
coolhandle01: This is the comment that actually matters and you're right on all three counts. Let me unpack each, because they're related but the fixes look different.

## (1) `programme_handle` threaded through every call

You're right - it's constant within a run, the agent should not be re-typing the handle for every Annotate Host / Finalise Recon / Probe Hostnames call. Each repetition is a token waste and, worse, a foot-gun: nothing today stops the agent from passing a *different* programme handle on a single call mid-run, which would silently corrupt scope checks for that one invocation. The schema is "exact HackerOne programme handle"; the workspace already has it; the tool should read it, not ask.

Concrete shape of the fix: the handle gets written into `sweep.json` at `Run Initial Sweep` time (it's already there indirectly via the `programme` block), and every downstream OA tool reads from the sweep / a singleton rather than taking it as a parameter. The wrapper signatures drop `programme_handle` entirely; the schemas lose that field.

This is **cross-agent**, not OSINT-specific. PM browses programmes by handle, DC submits by handle, TA's `List Programme Reports` takes a handle, VR's `Read Raw Finding` is per-run too. The same threading-through-every-call smell is going to land on #149 / #150 the moment I sweep them. Worth fixing at the squad level (`runtime.set_programme(handle)` + a `runtime.current_programme()` reader, similar to how `http.set_programme(...)` already works for the User-Agent header) rather than per-agent.

## (2) `scope_guard` as a `check_fn`, not a per-tool parameter

Also right. Today the scope filter is scattered inside `probe_hostnames_tool` and `detect_takeover_candidates_tool` as inline `filter_in_scope_impl([h...], programme)` calls. Each call site does the same dance; the parameter exists only because the dance needs it. That's exactly the shape `@pentest_tool(check_fn=...)` solved for the OWASP categories - the wrapper layer applies the cross-cutting policy, the body just runs the tool.

The right shape is something like:

```python
@cyber_tool(
    "Probe Hostnames",
    args_schema=_ProbeHostnamesArgs,
    scope_filter=filter_in_scope,  # check_fn applied to the hostnames input
)
def probe_hostnames_tool(hostnames: list[str]) -> list[Endpoint]:
    if not hostnames:
        return []
    return list(probe_endpoints_impl(hostnames))  # already scope-filtered
```

`@cyber_tool` (or a new `@scope_guarded` companion) runs the filter on a named input field before the body executes. Out-of-scope inputs get dropped before any DNS / HTTP fires - the guarantee is on the wrapper rather than relying on every body to remember the dance. PT wrappers that take `list[Endpoint]` from `recon.json` skip this because the endpoints were already scope-filtered at recon-finalise time; the guard fires on tools that accept arbitrary hostnames from the agent.

## (3) "OA may be overly simplified - serious refactor while the rules are allowed to change"

Yes. The act of writing per-field `Field(description=...)` for every parameter is exactly the audit that surfaces this - if you can't write a description for a parameter without sounding redundant or apologetic, the parameter probably shouldn't exist. Half the OSINT schemas I just wrote have descriptions that essentially say "this is the same thing again, sorry". That's a code smell the args_schema discipline made visible.

Concrete things I'd want to think about in an OA refactor (beyond the two above):
- `sweep_path: str = "sweep.json"` on every read tool - same problem as `programme_handle`. Workspace handle should be implicit when there's only one canonical sweep per run.
- `Annotate Host` has 7 parameters (3 required, 4 optional) - it's load-bearing but the shape is "everything the LLM might want to set". A typed `HostInsight`-shaped input would be cleaner, with the wrapper composing the auxiliary lookup (sweep, programme) itself.
- `lookup_cwe_tool` / `lookup_owasp_tool` are duplicated across OA, VR, TA with slightly different docstrings. Three near-identical wrappers is a smell.
- `LlmDetectionArgs` taking `list[Endpoint]` is the only one - all the other Endpoint-takers are on PT. Worth questioning whether OA should be passing endpoints around explicitly when it has a sweep file right there.

## Path forward

Three options:

**(A) Land this PR after the small fixes, file `feat(osint): API refactor` as a follow-up.** Args_schema discipline is independently valuable; the schemas get revised when the API gets cleaned up. Cheapest in flight time. Cost: some Field descriptions written here get rewritten in the refactor PR.

**(B) Hold this PR. Do the OA refactor first.** Then re-do the args_schema sweep on the cleaner API. Cleanest outcome but a lot more flight time and the args_schema work blocks on a bigger design conversation.

**(C) Land this PR but pause #149 / #150.** Use the OA refactor as the design template; once `programme_handle` threading and `scope_guard` get cleaner shapes, apply them to PM / DC / TA / VR in the same pass that adds their args_schemas. Avoids retrofitting the same smell across three more agents.

My instinct is **(C)** - the schemas this PR landed are net positive (the agent sees explicit fields where it didn't before), so reverting them is a step backward, but rolling forward into PM / DC blindly is going to bake `programme_handle` into three more agents at exactly the moment we agree it shouldn't be there. Pausing #149 / #150 gives the OA refactor room to set the pattern for the rest.

But I want your call on this. Happy to file the OA refactor issue with the three concrete pieces above (programme/sweep handle threading, scope_guard as check_fn / @decorator, broader API tightening) and pick a direction from there.

---

tests/test_osint_args_schemas.py
coolhandle01: Got it now - sorry, I misread "fixtures for endpoints" as "use the Endpoint fixture for the LlmDetection case" rather than "stop hardcoding example.com, use the URL fixtures the conftest provides". Fixed in `9406fa9`.

URL / hostname-taking accept-cases moved out of the parametrize into dedicated test methods that consume `victim_url`, `bystander_url`, and `callback_url`. Parametrize literals can't take fixtures, so the restructure was the only path. Each method names what role the fixture plays:

- `test_recon_open_ports_accepts_victim_host` - victim_url for in-scope hostname filter
- `test_cert_transparency_accepts_victim_apex` / `test_historical_urls_accepts_victim_apex` - victim_url apex via `urlparse(...).hostname.split(".", 1)[-1]`
- `test_probe_hostnames_accepts_victim_hostname` - victim_url for the hostnames list
- `test_detect_takeover_candidates_accepts_bystander_hostname` - **bystander_url** is the right intent here: takeover modelling is precisely "CNAME dangles to an out-of-scope target", which is what bystander_url exists for
- `test_annotate_host_accepts_victim_hostname` - victim_url for the annotated hostname

The Annotate Host rejection tests now derive their hostname from `victim_url` too. Added a `_annotate_host_base(hostname)` helper so the minimal-valid kwargs shape lives in one place and each rejection test stays focused on the field it exercises.

Parametrize accept-cases retained only for URL-less schemas (Run Initial Sweep, the recon-query slicers, Lookup CWE / OWASP, Uncovered Hosts, Finalise Recon) where there's no domain shape to be opaque about.

CI green: 1216 tests passing, 91.72% coverage, ruff / mypy / pylint / bandit all clean.

---

squad/osint_analyst/__init__.py
```diff
@@ -147,21 +272,61 @@ def historical_urls_tool(domain: str) -> list[str]:
...
+    # before invoking us; re-validate so we always see ``Endpoint`` instances
+    # regardless of whether the caller is the runtime or a direct test invocation.
+    parsed = [Endpoint.model_validate(e) for e in endpoints]
+    return [LlmEndpoint.model_validate(ep.model_dump()) for ep in detect_llm_endpoints(parsed)]
+
+
+class _ProbeHostnamesArgs(BaseModel):
+    """Explicit args_schema for the Probe Hostnames tool (#148)."""
+
+    hostnames: list[str] = Field(
```
coolhandle01: Could this be expressed as `list[Endpoint]`?

coolhandle01: Honest answer: semantically no, but there's a real tightening hidden in the question.

`Probe Hostnames` is the tool that *turns* hostnames into Endpoints - the input is pre-probe (we don't know status_code, technologies, parameters yet), the output is `list[Endpoint]` (what httpx came back with). Forcing the input to `list[Endpoint]` would mean the agent has to construct placeholder Endpoints with fake `status_code=0` / `technologies=[]` just to satisfy the type, then we throw all that away to re-probe. The shape would be lying about what it carries.

`Detect Takeover Candidates` is the same story - input is "resolve these via dnsx", output is `list[TakeoverCandidate]`. Pre-resolution, all we have is the hostname.

That said, `list[str]` is genuinely lax compared to the rest of the args_schema discipline. The tightening would be a `Hostname` typed-string (a Pydantic `Annotated[str, BeforeValidator(...)]` or a constrained-str subclass) that validates "this looks like a hostname, not a URL or garbage" - so `list[Hostname]` rejects `["https://foo.com"]` or `["not a host"]` upstream. That's a real gain: the agent occasionally hands us a URL when we asked for a hostname, and a typed wrapper would catch it before the scope filter does silently.

Cost vs scope: the `Hostname` model doesn't exist today; introducing it touches the model layer, both wrappers, and probably the underlying `filter_in_scope` / `probe_endpoints` helpers. That feels like exactly the kind of move the OA refactor we agreed to defer would naturally pull in - it's adjacent to the `Annotate Host` -> typed `HostInsight` input change and the broader API tightening.

So I'd leave `list[str]` here and add `Hostname` to the OA refactor issue's bullet list. Sound right, or do you want me to take a shot at `Hostname` in this PR?

---

squad/osint_analyst/__init__.py
coolhandle01: Got it - going with **(A) + the scope_guard piece as a focused follow-up**, not the full OA fidelity refactor. The "Frankensteiney" scope-guard plumbing is real but contained, and `@cyber_tool(scope_filter=...)` is small enough to be its own well-bounded PR.

Plan:

1. Land this PR as-is (the small fixes are in: StrEnums on role/priority, fixture-based test methods).
2. File a focused follow-up: `refactor(squad): @cyber_tool(scope_filter=...) check_fn for hostname-taking tools`. Scope is narrow - lift the inline `filter_in_scope_impl(...)` dance out of `probe_hostnames_tool` and `detect_takeover_candidates_tool` into the wrapper layer, with `programme` pulled from `runtime.current_programme()` (or equivalent) rather than threaded through the call. PT wrappers skip the guard because their `list[Endpoint]` already came scope-filtered out of recon-finalise.
3. Continue with #149 (PM + DC, 5 wrappers) and #150 (VR + TA + workspace) using the established pattern. Each will inherit the `programme_handle` smell on the input side, but you'd rather they ship with that smell visible than have me bake a half-baked workaround into three more agents.
4. The broader OA fidelity work - typed `Hostname`, `HostInsight`-shaped Annotate Host input, deduplicating CWE/OWASP lookups across agents, sweep_path implicit - stays in your hands per your message. Happy to file a tracking issue with the bullet list from my earlier comment as the audit trail, or leave it un-filed if you'd rather keep it in your head until you're ready.

Filing the scope_guard issue now. Let me know on the "tracking issue for the bigger OA refactor" question.

---

squad/osint_analyst/__init__.py
coolhandle01: I think Hostname probably could be quite useful throughout the codebase, *and* unlucky, you just sold it to me. It sounds like Hostname touches all of our "scope guard", so I think it comes under the issue this PR addresses in a way too. A useful kick in the shins if our problems before we run away and merge this PR.

coolhandle01: Sold. Landed in `5a0eac1`.

**Implementation**: `Hostname = Annotated[str, AfterValidator(_validate_hostname)]` in `models/__init__.py`. Strict RFC 1123 validation - rejects schemes (`://`), ports (`:`), paths (`/`), empty input, leading/trailing-hyphen labels, oversized labels (>63 chars), oversized total (>253 chars), garbage characters in labels (whitespace, newlines, etc.). Lowercases and strips on the way through, so downstream comparisons are case-insensitive without each call site re-lowering.

**Applied across six fields on five OSINT schemas**:

| Schema | Field |
|---|---|
| `_ProbeHostnamesArgs` | `hostnames: list[Hostname]` |
| `_DetectTakeoverCandidatesArgs` | `hostnames: list[Hostname]` |
| `_ReconOpenPortsArgs` | `host: Hostname \| None` |
| `_AnnotateHostArgs` | `hostname: Hostname` |
| `_CertTransparencyArgs` | `domain: Hostname` |
| `_HistoricalUrlsArgs` | `domain: Hostname` |

Wrapper signatures updated to match. Function bodies untouched - `Hostname` is `str` at runtime, so all existing string ops compose; the body's existing `.strip().lower()` calls are now redundant but idempotent and I left them as defensive code for the off-args_schema path.

**Tests**:

- `tests/test_models.py` `TestHostname` exhaustively covers the validator: apex / subdomain / single-label / numeric-label accept paths, lowercase / strip normalisation, and the full malformed set (empty, scheme, port, path, leading-or-trailing hyphen, empty label, oversized label, oversized total, whitespace in label, newline injection, non-string). Uses the `victim_url` fixture for the canonical accept case per the cybersquad-test-fixtures convention.
- `tests/test_osint_args_schemas.py` adds per-schema rejection cases (parametrized) for Hostname-typed fields plus an assertion that the lowercase normalisation flows through to the schema's parsed value.

CI green: 1235 tests passing (+19 from Hostname additions), 91.74% coverage, ruff / mypy / pylint 9.87 / bandit all clean.

Closes the "kick in the shins" - the scope guard now has a typed input edge instead of a `list[str]` that the scope filter quietly massages.

---

coolhandle01: Right - caught me leaving `example.com` literals in the Hostname tests. Fixed in `c06143f`.

Every accept / lowercase / strip / reject test in `TestHostname` now consumes `victim_url` and derives the hostname (and the malformed variants) via `urlparse(...).hostname`. The malformed-corpus parametrize collapsed into a single fixture-using method that loops the cases - parametrize literals cannot consume fixtures, so a single test method with a tuple list is the cleanest path. `localhost` and `10.0.0.1` accept cases keep their literals because no URL fixture exposes those shapes.

Same restructure on `test_hostname_field_rejects_malformed_input` in `test_osint_args_schemas.py` - each malformed input now built from the `victim_url`-derived hostname (`f"https://{host}"`, `f"{host}:8080"`, etc.). Test intent reads as "the in-scope target, deliberately mis-shaped" rather than opaque `example.com` literals.

CI: 1235 tests, 91.74% coverage, all linters clean. Same numbers as before - the restructure didn't change coverage.

---

Issue #149, opened 2026-05-22
refactor(h1): explicit args_schema on every HackerOne API tool


coolhandle01:
## Motivation

Follow-on from #143. The HackerOne-API tools on the Programme Manager and Disclosure Coordinator are the highest-stakes tool surface in the squad: a mis-call here is **publicly visible on hackerone.com** and cannot be silently retried. Wrong programme handle on `Submit Report` = a report filed against the wrong target = a public scope violation that the operator has to apologise for. Wrong filter on `Browse HackerOne Programmes` = the squad walks past the programme it should have picked.

The inferred schema cannot attach per-field descriptions; the explicit path can, and on these five tools the per-field guidance is doing real work - `handle` is the most-mis-called field type in the codebase, and the schema is the right place to spell out "exact programme handle as it appears in the URL, lowercase, no slashes".

## What changes

Every `@tool`-decorated function in `squad/programme_manager/__init__.py` and `squad/disclosure_coordinator/__init__.py` gets an explicit Pydantic `args_schema`:

- PM: `Browse HackerOne Programmes`, `Hydrate HackerOne Programme`, `Save Selected Programme`
- DC: `Submit Report`, `Check H1 Duplicate`

Field descriptions document:

- exact format expected (handle vs. URL, severity vs. CVSS, weakness ID vs. CWE string)
- which fields the H1 API treats as authoritative
- consequence of getting it wrong (irreversible for `Submit Report`)

## Constraints

- **No behaviour changes.** The H1 API surface is unchanged.
- **No new validation that the H1 API would not have applied.** The schema enforces shape (e.g. `handle: str` non-empty) and StrEnum-ish constraints where they exist; it does not duplicate H1's own validation.
- **Skill update.** `cybersquad-tool` (universal) gains the `args_schema` note if not landed by a sibling follow-up.
- **Tests follow.** Per-tool schema accept/reject tests, structurally aligned with `tests/test_pentest_args_schemas.py`. Tests must **not** call the H1 API - existing H1 tests are mocked and this stays the same.

## Scope

In: every `@tool` in `squad/programme_manager/__init__.py` and `squad/disclosure_coordinator/__init__.py`.

## Dependencies / ordering

Independent of the PT-cloud and OSINT follow-ups. The first follow-up to land owns the `cybersquad-tool` skill edit; the others inherit it.

## Acceptance criteria

- [ ] Every PM and DC `@tool` has an explicit `args_schema` Pydantic model.
[...]

PR #155, opened 2026-05-25
refactor(squad): explicit args_schema on triage, report, workspace, and PT recon tools (#150)

### LINKED ISSUE #150: refactor(squad): explicit args_schema on triage, report, and workspace tools

coolhandle01:
Closes #150.

Final-pass `args_schema` sweep: every cybersquad `@tool` wrapper now carries an explicit Pydantic schema; the signature-inferred path is unreachable. Completes the chain started in #143 / #146 (PT) and continued through #147 (PT cloud), #148 (OSINT), #149 (H1 API).

## What's in the diff

**The sweep itself** - 25 wrappers across VR (11), TA (7), workspace (3), PT recon/save (4). The `research_brief_tool` decorator takes `args_schema=` keyword-required, mirroring #143 / #146's `pentest_tool`. Per-field descriptions phrased as agent-facing targeting guidance; dedicated tests pin the CVSS-vector format and the Sanitise-Evidence reminder against quiet drift.

**Beyond the sweep** (review-driven, ~90% of the diff):

- **File splits**. PT (2034 lines) → orchestrator + 5 sub-modules + `probes/` and `cloud/` sub-packages. VR / OSINT similarly split. Public imports unchanged; only `mock.patch` targets in three test files moved.
- **Typed models**. `AuthoredAssessment` / `AuthoredDraft` (`models/triage.py`, `models/report.py`) as the LLM-authored half of Assess Raw Finding / Draft Vulnerability Report. `Finalise Research.items` now `list[AttackPlanItem]`, was `list[dict]`. `TakeoverCandidate.hostname` now `Hostname`. Per-agent shapes (`HostAnnotation`, `CWEEntry`, `OWASPEntry`, validation reports) moved to `models/`. Dropped two `R0913` and one `C0302` pylint suppression along the way.
- **`SquadTool` rename**. `@cyber_tool` now returns a thin `Tool` subclass exposing `args_schema` as a typed attribute; the brand is visible at every consumer site.
- **Pylint gate added** to CI (`fail-under = 9.85`, current 9.90). The contributor docs already required it; only the workflow didn't run it.
- **Fresh-eyes review feedback landed**: the two `programme_handle: str | None` schemas tightened to required, the orphan `models/primitives.py` FIXME paid down to design documentation, `models/attack.py` import layered through `models.primitives`, `AttackPlan*` re-exported from `models/__init__.py`.

## Test plan

- [x] `ruff check .` / `ruff format --check .` clean
- [x] `mypy . --ignore-missing-imports` clean (179 source files)
- [x] `pylint .` 9.90/10 (fail-under = 9.85)
- [x] `pytest -m unit --cov` - 1533 passed, 93.40% combined line + branch
- [x] `bandit -c pyproject.toml -r . -q` clean

## Out of scope (tracked in #156)

Cloud scope-of-target gate; `CvssVector` / `CweId` typed primitives; `programme_handle` workspace-state dedupe.


squad/penetration_tester/__init__.py
```diff
@@ -1764,7 +1763,30 @@ def jwt_check_tool(
     return list(check_jwt(token, endpoint, attacks))
 
 
-@tool("Recon Subdomains")
+class _PtReconSubdomainsArgs(BaseModel):
+    """Explicit args_schema for the PT's Recon Subdomains tool (#150)."""
+
+    recon_path: str = Field(
+        description=(
+            "Relative path to the OSINT Analyst's ``recon.json`` in the"
```
coolhandle01: Seems vague.. also, this file is way way too big to work with now

squad/penetration_tester/__init__.py
```diff
@@ -1776,7 +1798,63 @@ def recon_subdomains_tool(recon_path: str, host_filter: str | None = None) -> li
...
+    )
+    status: int | None = Field(
+        default=None,
+        description=(
+            "Conjunctive HTTP status filter (e.g. ``200`` returns only"
+            " endpoints that responded 200 during the sweep). Omit"
+            " (None) to skip the filter."
+        ),
+    )
+    tech: str | None = Field(
```
coolhandle01:
```suggestion
    #FIXME #83 this should be a list of Frameworks
    tech: str | None = Field(
```

squad/penetration_tester/__init__.py
```diff
@@ -1806,7 +1884,29 @@ def recon_endpoints_tool(
...
+    """Explicit args_schema for the PT's Recon Open Ports tool (#150)."""
+
+    recon_path: str = Field(
+        description=(
+            "Relative path to the OSINT Analyst's ``recon.json`` in the"
+            " current run directory. Same semantics as on ``Recon"
+            " Subdomains``."
+        ),
+    )
+    host: str | None = Field(
```
coolhandle01:
```suggestion
    host: Hostname | None = Field(
```

squad/technical_author/__init__.py
```diff
@@ -102,7 +201,112 @@ def list_programme_reports_tool(
...
+        description=(
+            "Developer-focused explanation of WHY the code is"
+            " vulnerable - the underlying class of flaw, not the"
+            " reproduction steps (those go in ``steps_to_reproduce``)."
+        ),
+    )
+    steps_to_reproduce: list[str] = Field(
+        description=(
+            "Numbered list a triager can follow verbatim. The validator"
+            " refuses steps under 10 characters - each step must be"
```
coolhandle01: We should provide "minimal steps to reproduce", so a numbered list of steps where each step is a documented command, so (oversimplified version):

1. Use the {tool} tool to show that {endpoint} exposes {framework} on {host}
    `some nmap command in scope`
    This yields the following response, visible in the attached evidence:
    `something from the evidence in the workspace that gets uploaded`
     Use the following command to see this in the attached files:
     `some jq or grep command or something`
2. Use the {attack} to show that {host} is vulnerable to {vulnerability}
    `somewhacky curl command`
     ...


Might need to make it so that the evidence/logs/recorded responses that are provided with the reports as supporting evidence are easily processed by tools that a triager will typically wish to use to verify our claims.

squad/technical_author/__init__.py
```diff
@@ -102,7 +201,112 @@ def list_programme_reports_tool(
...
+    )
+    remediation: str = Field(
+        description=(
+            "Actionable fix paired with an OWASP or CWE URL citation."
+            " The validator refuses remediations that name no fix or"
+            " carry no citation - use ``Lookup CWE`` / ``Lookup OWASP"
+            " Guidance`` to find the matching URL."
+        ),
+    )
+    cvss_vector: str = Field(
```
coolhandle01: Are there models now for cwe and cvss?

squad/vulnerability_researcher/__init__.py
```diff
@@ -130,7 +180,45 @@ def list_programme_reports_tool(
...
+    programme_handle: str = Field(
+        description=(
+            "Exact HackerOne programme handle the plan is drafted"
+            " against - stamped onto the AttackPlan and re-read by the"
+            " PT's ``Read Attack Plan``. Must match the handle the OSINT"
+            " Analyst recon was run against; a mismatch corrupts the"
+            " handoff."
+        ),
+    )
+    items: list[dict] = Field(
```
coolhandle01: `list[dict]` is to be avoided, I'm glad this PR surfaced it, there should be a strongly typed model for whatever this dict is

squad/vulnerability_researcher/__init__.py
```diff
@@ -250,7 +431,123 @@ def lookup_owasp_tool(query: str) -> list[OWASPEntry]:
...
+        description=(
+            "How this severity call relates to the PT's"
+            " ``severity_hint``: one of ``keep`` (the VR agrees with the"
+            " PT), ``raise`` (the VR's analysis warrants a higher band -"
+            " explain why in ``severity_rationale``), or ``lower``"
+            " (the VR's analysis warrants a lower band). The wrapper"
+            " coerces to ``SeverityDecision``; unknown values reject."
+        ),
+    )
+    severity: str = Field(
```
coolhandle01:
```suggestion
    severity: Severity = Field(
```

tests/test_pt_args_schemas.py
```diff
@@ -275,6 +292,30 @@ class TestSchemaAcceptReject:
...
+                    "recon_path": "recon.json",
+                    "status": 200,
+                    "tech": "wordpress",
+                    "host_contains": "admin",
+                    "offset": 0,
+                    "limit": 25,
+                },
+            ),
+            (_PtReconOpenPortsArgs, {"recon_path": "recon.json"}),
+            (_PtReconOpenPortsArgs, {"recon_path": "recon.json", "host": "api.example.com"}),
```
coolhandle01:
```suggestion
            (_PtReconOpenPortsArgs, {"recon_path": "recon.json", "host": f"api.{victim_url}"}),
```

squad/penetration_tester/__init__.py
coolhandle01: Tightened in 2da9bf9 - the description now names the field in `recon.json` the slicer reads (`subdomains`) and what kind of data the OA put there, mirroring the per-slicer specificity that the cloud / infra wrappers (`_S3CheckArgs` etc.) already carry. Did the same for `_PtReconEndpointsArgs.recon_path` (`endpoints`) and `_PtReconOpenPortsArgs.recon_path` (`open_ports`).

On the file size: agreed, this is pre-existing pain that #150 makes worse - the module is now 2034 lines (was 1922 pre-#150; my four wrappers added ~110 lines). Splitting `squad/penetration_tester/__init__.py` into a sub-package (probes / cloud / recon / save as separate modules, all re-exported through `__init__.py` so the `MEMBER.tools` registry stays intact) is structural and out of scope for this PR. Happy to open a follow-up issue and link it from here - want me to file one?

---

squad/technical_author/__init__.py
coolhandle01: Rewrote the `steps_to_reproduce` description in 2da9bf9 to ask for that exact shape: tool used, command/request, expected response excerpt, and a pointer to the supporting evidence. Example I put inline:

> Use sqlmap to confirm injection at /search?q. Run: `sqlmap -u 'https://victim.example.com/search?q=test' --batch`. The `[INFO]` line citing `boolean-based blind` appears in the response and is captured in the attached evidence (grep `boolean-based blind` in the sanitised log).

The deeper question - whether the supporting-evidence / replay-log handoff should become a workspace-typed artefact so triagers can replay verification commands against captured payloads rather than the live target - is bigger than #150 and probably ties into how the runtime crew stores per-finding evidence. Added a `FIXME` directly under the field flagging it as a follow-up rather than expanding scope here. Happy to file an issue if you want the framing tracked outside the FIXME.

---

squad/technical_author/__init__.py
coolhandle01: Not yet, no - good prompt to check. What exists today:

- **CWE**: `tools/cwe_data.CWEEntry` is the typed row (id / name / description / aliases / owasp_topic + computed MITRE URL), and `tools/cwe_data.lookup(query)` returns `list[CWEEntry]`. But there is no `CweId` primitive in `models.primitives` - the `cwe_id` field is a bare `int`, and the validator only catches mismatches at draft-validation time, not at args_schema time.
- **CVSS**: no model at all. `cvss_vector` is a bare `str`; `calculate_cvss_score(vector)` returns a `float`. The 3.1 vector format is documented in the field description (#150 puts it in the args_schema's `Field(description=...)`) but not type-enforced.

Both are real gaps in the typed primitive layer. Added FIXMEs on `_DraftReportArgs.cvss_vector` and `_DraftReportArgs.cwe_id` in 2da9bf9 flagging the follow-up. The shape would be:

```python
# models/primitives.py
CvssVector = Annotated[str, AfterValidator(_validate_cvss_3_1_vector)]
CweId = Annotated[int, AfterValidator(_validate_in_cwe_catalogue)]
```

With those, malformed vectors / unknown CWE ids reject at args_schema validation time rather than at score-compute / catalogue-lookup time. Want me to file the follow-up issue?

---

squad/penetration_tester/__init__.py
coolhandle01: Done in ffb8c1b - the PT, VR, and OSINT modules are split into per-responsibility sub-modules (since you wanted the file-size fix done as part of the final-pass framing rather than punted to a follow-up).

**Penetration Tester** (2034 → orchestrator + 5 sub-modules):
- `_decorator.py` (91) - `pentest_tool` decorator + `_PentestTool` Protocol + `_parse_endpoints` / `_recon_from_path` adapters
- `probes.py` (1188) - the 25 `@pentest_tool` wrappers
- `cloud.py` (528) - the 18 `@cyber_tool` cloud / infra wrappers
- `recon.py` (182) - the three typed `recon.json` slicers
- `findings.py` (66) - `Save Findings`
- `__init__.py` (306) - imports, `MEMBER` assembly, re-exports

**Vulnerability Researcher** (779 → orchestrator + 3 sub-modules):
- `_decorator.py` (57) - `research_brief_tool` + `_ResearchBriefTool` Protocol
- `research.py` (200) - the three research-task tools
- `triage.py` (551) - the eight triage-task tools + `_parse_severity` / `_load_programme` helpers
- `__init__.py` (119) - orchestrator

**OSINT Analyst** (671 → orchestrator + 3 sub-modules):
- `_helpers.py` - shared `_load_programme` loader
- `discovery.py` (394) - sweep + slicers + passive expansion + active hostname probes + takeover detection
- `curation.py` (255) - lookup + `Annotate Host` + `Uncovered Hosts` + `Finalise Recon`
- `__init__.py` (119) - orchestrator

Every public name (`MEMBER`, every wrapper, every `_*Args` schema class) still imports from the same `squad.<agent>` path, so `crew.py`, `tasks.py`, and the args_schema contract tests don't need to change. Three test files (`test_squad_{penetration_tester,vulnerability_researcher,osint_analyst}.py`) updated their `mock.patch` targets to point at the canonical submodule path - `mock.patch` resolves names where they are looked up, and the lookup site moved when the wrappers moved.

`probes.py` and `cloud.py` carry an explicit `# pylint: disable=C0302` with the canonical-registry rationale (each is one cohesive responsibility - the per-probe / per-engine contract is the whole point of the file). `triage.py` carries the same for the same reason - it's the long `Assess Raw Finding` schema that dominates the line count, not redundant code.

CI: ruff clean, ruff format clean, mypy clean (161 source files), pytest 1508 unit tests pass at 91.90% combined line + branch coverage, bandit clean, pylint 9.94/10 (three pre-existing R0913/R0914/R0917 findings on functions that already exceeded the cap before #150).

---

squad/osint_analyst/_helpers.py
```diff
@@ -0,0 +1,23 @@
...
+Sweep``) and ``curation.py`` (``Finalise Recon``) can call it without
+the two modules cross-importing.
+"""
+
+from models.h1 import Programme
+from tools import http
+from tools.h1_api import h1
+
+
+def _load_programme(programme_handle: str | None) -> Programme:
```
coolhandle01: 100% this should be in workspace tools if anywhere

squad/osint_analyst/curation.py
```diff
@@ -0,0 +1,255 @@
...
+
+from models import HostInsight, Hostname, HostPriority, HostRole
+from squad import cyber_tool
+from squad.osint_analyst._helpers import _load_programme
+from tools.cwe_data import CWEEntry
+from tools.cwe_data import lookup as cwe_lookup
+from tools.owasp_data import OWASPEntry
+from tools.owasp_data import lookup as owasp_lookup
+from tools.recon_insights import (
+    HostAnnotation,
```
coolhandle01: HostAnnotation and ReconFinalisationError look like they should come from models/...

coolhandle01: So do CWEEntry and OWASPEntry

squad/osint_analyst/discovery.py
```diff
@@ -0,0 +1,394 @@
...
+from tools.h1_api import h1
+from tools.recon import (
+    cert_transparency,
+    detect_llm_endpoints,
+    detect_takeover_candidates,
+    historical_urls,
+    run_recon,
+)
+from tools.recon import probe_endpoints as probe_endpoints_impl
+from tools.recon.dnsx import TakeoverCandidate
```
coolhandle01: TakeoverCandidate should probably come from models/... ?

squad/osint_analyst/discovery.py
```diff
@@ -0,0 +1,394 @@
...
+            " (lowercase, no slashes, no spaces). The sweep runs subfinder,"
+            " httpx, nmap, ffuf, and passive TLS / DNS checks against the"
+            " programme's structured scope - the handle is the authoritative"
+            " key the H1 API uses to look the scope up."
+        ),
+    )
+
+
+@cyber_tool("Run Initial Sweep", args_schema=_RunInitialSweepArgs)
+def run_initial_sweep_tool(programme_handle: str) -> str:
```
coolhandle01, line 53: Programme == rundir, no? This is really bugging me haha

squad/osint_analyst/discovery.py
```diff
@@ -0,0 +1,394 @@
...
+            " regardless of position). Omit or pass null to get the full"
+            " list. Useful for scoping a per-role annotation pass."
+        ),
+    )
+
+
+@cyber_tool("Recon Subdomains", args_schema=_ReconSubdomainsArgs)
+def recon_subdomains_tool(
+    sweep_path: str = "sweep.json", host_filter: str | None = None
+) -> list[str]:
```
coolhandle01: Could this be expressed as list[Hostname] ? Or no?

squad/penetration_tester/_decorator.py
```diff
@@ -0,0 +1,90 @@
...
+every wrapper file.
+"""
+
+from collections.abc import Callable
+from typing import Protocol, cast
+
+from pydantic import BaseModel
+
+from models import Endpoint, RawFinding, ReconResult
+from squad import CrewAITool, cyber_tool
```
coolhandle01: It's weird seeing "CrewAI" coming from squad, if that's ours we should name it better.. if you'd name it CyberTool, we might have questions to ask ourselves

squad/penetration_tester/cloud.py
```diff
@@ -0,0 +1,528 @@
...
+        description=(
+            "Relative path to recon.json in the run directory. Buckets are"
+            " derived from the programme handle and any S3 subdomains the"
+            " OSINT Analyst surfaced. Worth firing when the target is known"
+            " to use AWS or when *.s3 / *.s3.amazonaws.com subdomains appear"
+            " in recon."
+        ),
+    )
+
+
```
coolhandle01: Ya know, I don't like these cloud checks we have that try to guess at storage or db resource URLs. Seems risky. Please add a `#fixme` to note that checks against cloud resources should be applied only when a cloud resource URL is definitely explicitly in scope, or the squad is working on that cloud providers program should such a program exist and be chosen by the PM

squad/penetration_tester/cloud.py
```diff
@@ -0,0 +1,528 @@
...
+            "Relative path to recon.json in the run directory. Fire when"
+            " open_ports shows 5432. CRITICAL if trust authentication"
+            " allows connection without a password; MEDIUM if the port is"
+            " exposed but credentials are required (unnecessary internet"
+            " exposure)."
+        ),
+    )
+
+
+@cyber_tool("Exposed PostgreSQL Check", args_schema=_PostgresqlCheckArgs)
```
coolhandle01: Since this files is so long we should extract some stuff I would start with exposed DB checks

squad/penetration_tester/cloud.py
coolhandle01: Maybe instead of recon_path it should be Hostname or Endpoint in some cases? 😬

squad/osint_analyst/discovery.py
coolhandle01: Acknowledging - this is a real design tension. You're right that the run directory already implies the programme (PM's `Save Selected Programme` sets `runtime.programme_handle` and `runtime.run_dir()` is keyed by it), so every wrapper that takes `programme_handle: str` is duplicating workspace state and creating an extra place the LLM can mis-thread it. Same pattern shows up on `Finalise Recon`, `Run Initial Sweep`, `Assess Raw Finding`, `Finalise Triage`, `List Programme Reports`, `Finalise Reports`.

The structural fix is making `load_programme()` read `runtime.programme_handle` by default (with the parameter as override), and dropping `programme_handle` from every args_schema where it's redundant. That's a meaningful API simplification touching ~7 tools and their tests, and it changes the agent-facing contract for those tools - probably wants its own PR with a focused "trust runtime state" framing rather than landing as the tail of #150.

Want me to file a follow-up? Easy yes; happy to scope it now and link the issue back here.

---

squad/penetration_tester/cloud.py
coolhandle01: FIXME added in 9fa5756 - per-sub-module in the new `cloud/` package (`storage.py`, `databases.py`, `panels.py`, `dashboards.py`, `service_discovery.py`, plus the package-level docstring) flagging the scope-of-target risk you raised. The recurring text:

> these checks derive bucket / container / host:port pairs from the programme handle and any service-shaped subdomains the OSINT Analyst surfaced. They should only fire when a cloud resource URL is definitely explicitly in scope (the programme's structured scope names the resource), or when the squad is working on a programme run by that cloud provider itself.

On the `recon_path` -> `Hostname` / `Endpoint` direction: agreed in principle - the LLM should pick the target explicitly rather than the wrapper deriving "anything S3-shaped from recon". But that's an API change for 18 wrappers (+ their tests + the agent prose teaching the new shape), and it ties into the structural fix above (a `Programme`-aware pre-flight gate that refuses cloud probes against not-in-scope hosts). Same follow-up as the FIXMEs reference. Happy to file the issue now if you want it tracked outside the per-file FIXMEs.

---

coolhandle01 (commented):
## Fresh-eyes review

### TL;DR

Mechanically excellent - all CI green (ruff/format/mypy/pylint 9.90/pytest 1533 passed/bandit), coverage 93.4%. The args_schema discipline is real: closed-world contract tests for PT, VR, TA mean a new typed tool without a registry entry fails CI before review. The MCP-integration runway you are building toward (typed `Hostname` / `HttpUrl` primitives, per-field `Field(description=...)` as agent-facing targeting guidance, schema-level reject upstream of HTTP requests) is sound.

The big concerns: two `programme_handle` schemas declare `str | None` but the description says "Required" - exactly the gap args_schema was supposed to close - and the skills have one drift each worth fixing now.

---

### Security / correctness

#### 1. Required-but-typed-optional `programme_handle` defeats args_schema (HIGH-priority for the PR's stated goal)

Two fields declare `str | None = None` but the description says "Required":

`squad/vulnerability_researcher/triage.py:254-262` - `_AssessRawFindingArgs.programme_handle`:
```python
programme_handle: str | None = Field(
    default=None,
    description=(
        "Exact HackerOne programme handle the scope / floor gates"
        " check against. Required - the gate refuses out-of-scope"
        ...
    ),
)
```

`squad/osint_analyst/curation.py:139-146` - `_AnnotateHostArgs.programme_handle` (same pattern, description says "Required - ... `None` raises ValueError").

The LLM reads the JSON schema (where it is optional) and the description (where prose says required). Schema validation passes for `{}`, then `_load_programme(None)` raises at runtime - the exact "mis-call costs a request / corrupts the artefact" path this PR's discipline is supposed to close at args_schema time. This is the only place in the PR where the discipline itself leaks into the merged code.

Fix is one line each: `programme_handle: str = Field(description=...)` (no default). #156 talks about dropping this parameter entirely; until that lands, the current schema lies about its own contract.

#### 2. Cloud scope-of-target gap (already tracked in #156)

The package-level FIXME in `squad/penetration_tester/cloud/__init__.py:21` correctly flags this: every cloud check (S3, ES, Mongo, Grafana, ...) derives its target from `recon.json` rather than from an explicit agent pick. If the OSINT sweep picked up a third-party S3 bucket that happens to share a name with the programme, the probe fires against the third party. The args_schema discipline here gives no protection - `recon_path: str` is just a filename, not a typed scope-gated target. Tracked in #156, right call - worth calling out in the PR body so reviewers know this PR doesn't close that risk and isn't trying to.

#### 3. Hostname primitive: IPv6 silently rejected

`models/primitives.py` `_validate_endpoint_url` runs the URL's parsed hostname through `_validate_hostname`. For IPv4 (`1.2.3.4`) the per-label regex passes. For IPv6 (`[2001:db8::1]`), `parsed.hostname` returns `2001:db8::1` which contains `:` and rejects with "hostname must not include a port". Probably fine for HackerOne programmes that target DNS-bound assets, but worth a one-line note in the validator's docstring so a future contributor isn't surprised when a programme lists IPv6-bound assets in scope.

---

### Clean code

#### 4. Redundant normalisation now that fields are typed

`squad/osint_analyst/curation.py:183` - `annotate_host_tool` body:
```python
insight = HostInsight(
    hostname=hostname.strip().lower(),  # Hostname validator already strips+lowers
    ...
)
```

`Hostname`'s validator does `value.strip().lower()` then validates. The `hostname` parameter is typed `Hostname` so by the time the wrapper body runs, it is already cleaned. Same for `discovery.py:340` and `discovery.py:393` (`[h.strip().lower() for h in hostnames if h.strip()]`) - if those lists are typed `list[Hostname]`, the strip/lower is dead code. Low priority.

#### 5. Repeated `recon.json` parse cost

Every cloud wrapper calls `_recon_from_path(recon_path)` which reads and parses the full ReconResult (~115KB / 30K tokens per the cybersquad-tool skill). A session running all 18 cloud probes re-parses recon 18 times. Not a correctness issue and not worth fixing if cloud tools never co-run, but worth flagging - a cheap LRU cache keyed on `(recon_path, mtime)` inside `_recon_from_path` would be a four-line win.

#### 6. Cloud DB check duplication

`squad/penetration_tester/cloud/databases.py` has 6 nearly-identical wrappers (`elasticsearch_tool`, `couchdb_tool`, `redis_tool`, `mongodb_tool`, `postgresql_tool`, `mysql_tool`), each:
```python
recon = _recon_from_path(recon_path)
findings = []
for host, ports in recon.open_ports.items():
    if PORT in ports:
        findings.extend(check_X(host))
return list(findings)
```

A helper `_port_gated_check(recon_path, port, check_fn) -> list[RawFinding]` would collapse these to one line each. Out of scope per the PR body's deliberate "consolidating duplicated lookups is a separate refactor" line - worth a tracking issue if there isn't already one.

#### 7. FIXME grammar drift

CONTRIBUTING.md prescribes `# FIXME: ... - tracked in #NN`. Current codebase mixes:
- `# FIXME #83 this should be a list of Frameworks` - no `:`
- `# FIXME(#152 follow-up):` - parenthesised
- `# FIXME #156:` - colon, no parens
- `# FIXME: migrate to Pydantic's built-in HttpUrl ...` (`models/primitives.py:82`) - **no issue number at all**, which violates the CONTRIBUTING rule

The orphaned FIXME at `models/primitives.py:82` is worth fixing in this PR - either pin it to #156 (which mentions `CvssVector` / `CweId` primitives but not the `HttpUrl` migration) or open a new tracking issue. The format inconsistency is a style sweep for another PR.

#### 8. `from models import Severity` inside `models/attack.py`

`models/attack.py:16` does `from models import Severity` instead of `from models.primitives import Severity`. The `cybersquad-tool` skill says "modules import only from layers below them ... inside the package, prefer the per-module path so circular-import dances stay out of reach." This works today only because `models.attack` is never imported from `models/__init__.py` - consumers all `from models.attack import ...`. Brittle; one line to fix.

---

### Skill notes

The skills are **mostly current** - `cybersquad-tool` was the right place to land the universal `args_schema` rule in #149 and this PR audits without changing it. Two small drifts:

#### Drift A: re-export claim does not match `models/__init__.py`

`cybersquad-tool` says "Consumers can still `from models import X` (re-exports preserve the public surface)" - but `AttackPlan`, `AttackPlanItem`, `AttackPlanFinalisationError` are documented in `models/__init__.py` but **not in `__all__`** and not re-exported. Every consumer uses `from models.attack import ...`. Pick one:
- Add to the re-exports (matches the skill text, but ratifies the brittle import in `models/attack.py:16`).
- Update the skill: "Some new domain modules carry shapes that are not re-exported; import from `models.<sub>` directly. The `__all__` in `models/__init__.py` is the source of truth."

#### Drift B: `cybersquad-tool` is getting big

199 lines including a substantial "Upstream alignment" section explaining the divergence from CrewAI's `output_pydantic` advice. That is reference-grade material - useful, but the skill loads in full on the first matching edit. Two options worth considering when this skill next needs an edit:

1. **Extract `cybersquad-models`** (triggers on `models/*.py`). Pull the "Typed string primitives" + "Where models live" sections out. Lets `cybersquad-tool` focus on the `@cyber_tool` / `@pentest_tool` / `@research_brief_tool` wrapper conventions; edits to `models/*.py` then load the relevant skill.
2. **Move "Upstream alignment" + "Divergence 1 & 2" to a `cybersquad-tool/references/divergence-from-crewai.md`** loaded only on demand. Standard `skill-creator` progressive-disclosure pattern, but the contributor skills currently do not use it (per `cybersquad-skill`: "cybersquad runtime skills are progressive-disclosure markdown only - no `scripts/` or `references/` directories"). That rule is correct for runtime skills; revisit for contributor skills if this one keeps growing.

Neither is urgent. Flagging in case the next iteration is the right moment.

#### Layering observation

The three per-agent `test_args_schemas.py` files (PT 476 lines, VR 384 lines, TA 308 lines) are now structurally identical: `_<AGENT>_SCHEMAS` mapping, three contract tests (wires-schema, fields-have-descriptions, closed-world-mapping), then per-domain accept/reject parametrize. A shared `tests/squad/_args_schema_helpers.py` exposing an `assert_agent_schemas_match(member, mapping)` function would collapse the contract trio to three lines per agent. The per-domain accept/reject stays per-file because the schemas differ. Worth a tracking issue if a fourth agent's args_schema test is on the horizon.

---

### What's good

- **Closed-world `_<AGENT>_SCHEMAS` mappings** with the private-prefix discovery check are the right design. A new typed tool without a mapping entry fails CI before review - the test doing exactly the job the skill text says it does.
- **Per-field descriptions phrased as targeting guidance** ("fire when open_ports shows 9200", "skip endpoints without parameters") not type info - exactly what `cybersquad-tool` calls for. The CVSS vector format pinned in the field rather than the docstring is a clear win.
- **`AuthoredAssessment` / `AuthoredDraft` as the LLM-authored half** of the args_schema is a clean separation. The wrapper-side carry-forward fields (target / vuln_class / severity_hint) cannot be contradicted by the LLM - a real correctness guarantee.
- **Empty `_ListRunFilesArgs` / `_ReadAttackPlanArgs` declared anyway** so the closed-world check covers them - small thing, but a less disciplined codebase would have skipped these.
- **`pentest_tool` and `research_brief_tool` composing on top of `cyber_tool`** rather than each implementing its own arg-schema plumbing is the right abstraction. Cheaper to maintain when MCP-specific decoration lands.

---

### For the MCP integration runway

The current shape is in good condition for it:
- The `LlmEndpoint` -> `DiscoveredMCP` FIXME in `models/asset.py:120` is the right place for it to land - `models/framework.py` (per #83) as the home, with the advertised tool inventory + suspicious-docstring flags + schema-laxness markers.
- `@cyber_tool` already accepts a generic `args_schema=` so an MCP-discovered tool wrapper can compose on top exactly the same way `pentest_tool` does. The `Hostname` / `HttpUrl` typed primitives reject scheme/port/path mismatches upstream of any MCP `initialize` call - that's the SSRF/scope-bypass risk class handled.
- The biggest gap for MCPs specifically: MCP servers advertise tool schemas, and the agent reasons about *those* schemas, but right now no harness re-types an MCP-advertised tool through `@cyber_tool` to inherit cybersquad's per-field-description discipline. Substantial new work (probably its own issue), but the foundation here doesn't paint into a corner.

---

### Recommended actions before merge

1. **Fix the two `programme_handle: str | None` schemas** (or document why None is allowed and the runtime ValueError is the contract). This is the only place in the PR where the args_schema discipline itself leaks. One-line each.
2. **Fix `models/primitives.py:82` FIXME to reference an issue** (#156 if it covers the migration, otherwise new).
3. **Update the PR body** to enumerate the file splits / new typed models / pylint gate so reviewers see what they are looking at - the body currently describes the args_schema sweep, which is roughly 10% of the diff.
4. **Optional**: fix `models/attack.py:16` import to use `models.primitives`.

Everything else can land as follow-ups against #156 or as standalone tracking issues.

---

coolhandle01:
### One more thing - the completionist framing

Worth saying out loud: this PR is the **closing chapter** of the args_schema journey - #143 -> #146 -> #147 -> #148 -> #149 -> #150. Every `@tool` wrapper in the squad now carries an explicit Pydantic contract; the signature-inferred path is unreachable; the closed-world contract tests refuse a new typed tool that skips the registry. That's the discipline landing, end-to-end.

Which means the loose ends I flagged are not "follow-ups". They are the dust on a finished piece of work. Specifically:

- The two `programme_handle: str | None` schemas (`_AssessRawFindingArgs`, `_AnnotateHostArgs`) are the **only** places in the whole squad where the args_schema lies about its own contract. Every other field across PT/VR/TA/OSINT/PM/DC/H1 is honest. Two one-line fixes and the discipline is uniformly true, not "true except here and here".
- The orphan FIXME at `models/primitives.py:82` is the **only** un-tracked FIXME in the diff. Pin it to an issue and CONTRIBUTING.md's grammar holds across the whole file tree.
- The `from models import Severity` in `models/attack.py:16` is the **only** intra-package import that bypasses the layer rule the skill teaches. One line.
- The `AttackPlan*` re-export gap is the **only** place where the skill's "from models import X works" claim does not match `__all__`. Either add the three names or update the one sentence.

Four things. None bigger than a line or two. With them landed, the args_schema sweep is not "mostly done with known caveats" - it is *done*, the skill matches the code matches the tests, and the next contributor reads a codebase that is honest about what it is. That is the right closing posture before MCP integration starts adding a new layer on top - the foundation should not carry "we'll get to those" notes into the next chapter.

Prim and proper. Let's land it that way.

---

Issue #156, opened 2026-05-25
refactor: post-#150 typed-primitive gaps + cloud scope gate + programme-handle dedupe


coolhandle01:
## Motivation

`#150` (the universal `args_schema` sweep) landed a handful of `FIXME`s pointing at typed-primitive gaps and a structural scope-of-target concern that surfaced during review. They are real but each is bigger than the PR they surfaced in, so they got `FIXME`'d rather than expanded scope. Per `CONTRIBUTING.md`'s FIXME grammar, every `FIXME` should reference a tracking issue; this is that issue.

### Typed primitives for CVSS and CWE

- `models/primitives` carries `Hostname` and `HttpUrl` but no `CvssVector` or `CweId`. Both are bare `str` / `int` on the args_schema and the validator only catches mismatches at score-compute / catalogue-lookup time.
- Sketch:

  ```python
  # models/primitives.py
  CvssVector = Annotated[str, AfterValidator(_validate_cvss_3_1_vector)]
  CweId = Annotated[int, AfterValidator(_validate_in_cwe_catalogue)]
  ```

- Drop the `FIXME`s on `_DraftReportArgs.cvss_vector` / `cwe_id` (`squad/technical_author/__init__.py`); same wins apply to the `cvss_vector` field on every other args_schema that carries one (`_VrCalculateCvssArgs`, `_TaCalculateCvssArgs`, `_AssessRawFindingArgs`).

### Scope-of-target gate on cloud / infra wrappers

- Every `@cyber_tool` under `squad/penetration_tester/cloud/` derives its target (bucket name, hostname, port pair) from `recon.json` rather than from the agent's explicit pick. Risk: probing third-party tenants that happen to share a bucket name or hostname pattern with the in-scope programme.
- Proposed shape: a `Programme`-aware pre-flight gate that refuses any cloud probe against a target not covered by the programme's structured scope. Lifts the per-sub-module `FIXME` in `cloud/__init__.py`.
- Possibly couples with replacing `recon_path: str` parameters on cloud wrappers with typed `Hostname` / `Endpoint` picks so the LLM explicitly names the target rather than the wrapper deriving "anything S3-shaped from recon".

### Workspace-typed supporting-evidence artefact

[...]

Issue #153, opened 2026-05-24
refactor(squad): @cyber_tool(scope_filter=...) check_fn for hostname-taking tools


coolhandle01:
## Motivation

Surfaced during review of #152. The OA agent has two tools that take arbitrary hostnames from the agent and run external probes against them (`probe_hostnames_tool`, `detect_takeover_candidates_tool`). Both wrap an inline scope filter today:

```python
@cyber_tool("Probe Hostnames", args_schema=_ProbeHostnamesArgs)
def probe_hostnames_tool(hostnames: list[str], programme_handle: str) -> list[Endpoint]:
    if not hostnames:
        return []
    programme = _load_programme(programme_handle)
    in_scope = filter_in_scope_impl([h.strip().lower() for h in hostnames if h.strip()], programme)
    if not in_scope:
        return []
    return list(probe_endpoints_impl(in_scope))
```

Two smells the args_schema sweep made visible:

1. **`programme_handle` is a per-call parameter** for a value that is constant within a run. The schema for it reads "Exact HackerOne programme handle. Required - the scope guard in ``filter_in_scope`` needs a real Programme to check against" - the parameter only exists because the dance needs it. The agent could in principle pass a *different* handle on a single call mid-run and silently corrupt scope for that invocation.
2. **The dance is duplicated across call sites**, and the contract "out-of-scope inputs are dropped before any external request fires" lives in two function bodies rather than at the wrapper layer where it can be enforced.

This is the exact shape that `@pentest_tool(check_fn=...)` solved for OWASP categorisation: the wrapper layer applies the cross-cutting policy, the body just runs the tool. Lifting it out of the bodies makes the guarantee explicit at the wrapper site rather than relying on every body to remember.

## What changes

Extend `@cyber_tool` (in `squad/__init__.py`) to accept an optional `scope_filter` parameter. When set, the wrapper inserts a pre-call step that runs the filter against a named input field before the body executes:

```python
@cyber_tool(
    "Probe Hostnames",
    args_schema=_ProbeHostnamesArgs,
    scope_filter=("hostnames", filter_in_scope),
)
def probe_hostnames_tool(hostnames: list[str]) -> list[Endpoint]:
    if not hostnames:
        return []
    return list(probe_endpoints_impl(hostnames))  # already scope-filtered
```

Plus the matching `runtime.current_programme()` reader so the filter has the Programme without the wrapper signature carrying it:

```python
# in squad/__init__.py or runtime.py
[...]

PR #158, opened 2026-05-25
refactor(squad): wrapper-level scope guard via @cyber_tool(scope_filter=...) (#153)

coolhandle01:
## Summary

Closes the design half of #153 that survives once #156 absorbs the broader `programme_handle` dedupe across the squad. Every "agent picked a target, run a scan / attack against it" tool can now declare its scope guard on the decorator instead of repeating the dance in every body.

- **Mechanism** (`squad/__init__.py`): `@cyber_tool` takes an optional `scope_filter=(field_name, filter_fn)`. When set, the wrapper validates args via `args_schema`, looks up the named field, and - if non-empty - calls `filter_fn(values, current_programme())` and rewrites the field before the body runs. Positional and keyword call sites are both honoured via `inspect.signature.bind_partial`; the `current_programme` lookup is lazy to dodge the `squad.workspace_tools` -> `squad` import cycle.
- **Loader** (`squad/workspace_tools.py`): `load_programme(programme_handle)` (which hit the H1 API every call) replaced by `current_programme()` which reads `<run_dir>/programme.json` - the snapshot the PM's `Save Selected Programme` already writes at run start. No more redundant H1 calls; the workspace is the contract.
- **First application** (`squad/osint_analyst/discovery.py`): `probe_hostnames_tool` and `detect_takeover_candidates_tool` use `scope_filter=("hostnames", _normalise_and_filter_hostnames)`. `programme_handle` dropped from both args_schemas and signatures (per #153 acceptance criteria). The shared helper lives at module scope so both wrappers normalise + filter exactly once.
- **Pending #156**: the other five `load_programme()` callers (`annotate_host_tool`, `finalise_recon_tool`, `assess_finding_tool`, `finalise_triage_tool`) switch to `current_programme()` in their bodies and carry a `FIXME(#156)` on the now-unused `programme_handle` parameter. #156 is the PR that drops the field from their args_schemas in one sweep.
- **Skill**: `.claude/skills/cybersquad-tool/SKILL.md` gains a "Scope guards belong on the wrapper" section with the canonical shape, plus two new anti-patterns (inline scope dance in a tool body; `programme_handle` field on schemas where it duplicates workspace state).

## Design notes

- `current_programme()` is no-arg by design (per discussion on #153 - the run dir already implies the programme via `runtime.programme_handle`). Sourcing from `<run_dir>/programme.json` rather than `sweep.json` keeps recon-finalisation orthogonal to scope guards.
- The mechanism is generic, but only the two OSINT probes that already had an inline scope dance are converted here. The bigger "tools that take a `recon_path` should take typed picks instead" surface (which the wrapper enables) is deliberately deferred to a follow-up - the parameter-shape refactor is a separate slice from introducing the scope guard.
- The PT cloud pre-flight gate is left to #156's third bullet - it has a different shape (refuses rather than filters) and the same wrapper can grow that pattern once the filter pattern is bedded in.

## Test plan

- [x] `ruff check .` + `ruff format --check .` clean
- [x] `mypy . --ignore-missing-imports` clean (179 source files)
- [x] `pylint .` at 9.90/10 (floor 9.85); no new R0913 / R0917
- [x] `pytest -m unit --cov --cov-report=term-missing`: 1536 passing, 93.43% combined line + branch coverage (floor 90%)
- [x] `bandit -c pyproject.toml -r . -q` - only pre-existing nosec warning
- [x] New `TestCurrentProgramme` covers the run-dir read path + the missing-file `FileNotFoundError` branch
- [x] New `TestCyberToolScopeFilter` covers the filter-runs-before-body path + the empty-input short-circuit (asserts `current_programme()` is not reached)
- [x] OSINT probe "drops out of scope" tests now exercise the real `filter_in_scope` against the fixture programme end-to-end (no more closure-captured mocks)

---
_Generated by [Claude Code](https://claude.ai/code/session_012ZtVnw7gfEcLPZKtmsTqnE)_

### DIFF
--- a/.claude/skills/cybersquad-tool/SKILL.md
+++ b/.claude/skills/cybersquad-tool/SKILL.md
@@ -49,7 +49,6 @@ Any tool that takes an agent-supplied target (hostname, endpoint, URL) and runs 
 
 ```python
 from squad import cyber_tool
-from squad.workspace_tools import current_programme  # noqa: F401  (loaded lazily)
 
 def _normalise_and_filter_hostnames(
     hostnames: list[Hostname], programme: Programme
--- a/.claude/skills/cybersquad-tool/SKILL.md
+++ b/.claude/skills/cybersquad-tool/SKILL.md
@@ -43,6 +43,43 @@ Rules:
 
 Specialist wrappers compose on top of `@cyber_tool`: `@pentest_tool` (pentest probes - layers OWASP categories from `check_fn.owasp_categories` into the docstring) and `@research_brief_tool` (Vulnerability Researcher - layers `PROBE_VOCABULARY` and `RECON_EVIDENCE_KINDS`) both call `cyber_tool` underneath and pass `args_schema=` through.
 
+### Scope guards belong on the wrapper, not in the body
+
+Any tool that takes an agent-supplied target (hostname, endpoint, URL) and runs an external scan / attack against it declares the guard on the decorator, not inline in the body:
+
+```python
+from squad import cyber_tool
+from squad.workspace_tools import current_programme  # noqa: F401  (loaded lazily)
+
+def _normalise_and_filter_hostnames(
+    hostnames: list[Hostname], programme: Programme
+) -> list[Hostname]:
+    cleaned = [h.strip().lower() for h in hostnames if h.strip()]
+    return filter_in_scope(cleaned, programme)
+
+
+@cyber_tool(
+    "Probe Hostnames",
+    args_schema=_ProbeHostnamesArgs,
+    scope_filter=("hostnames", _normalise_and_filter_hostnames),
+)
+def probe_hostnames_tool(hostnames: list[Hostname]) -> list[Endpoint]:
+    """The wrapper has already dropped out-of-scope hostnames; the body
+    just runs the probe."""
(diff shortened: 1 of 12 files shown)


.claude/skills/cybersquad-tool/SKILL.md
```diff
@@ -43,6 +43,43 @@ Rules:
 
 Specialist wrappers compose on top of `@cyber_tool`: `@pentest_tool` (pentest probes - layers OWASP categories from `check_fn.owasp_categories` into the docstring) and `@research_brief_tool` (Vulnerability Researcher - layers `PROBE_VOCABULARY` and `RECON_EVIDENCE_KINDS`) both call `cyber_tool` underneath and pass `args_schema=` through.
 
+### Scope guards belong on the wrapper, not in the body
+
+Any tool that takes an agent-supplied target (hostname, endpoint, URL) and runs an external scan / attack against it declares the guard on the decorator, not inline in the body:
+
+```python
+from squad import cyber_tool
+from squad.workspace_tools import current_programme  # noqa: F401  (loaded lazily)
```
coolhandle01: Hmm I don't like building this skill on a noqa..

Issue #76, opened 2026-05-15
XXE via OOXML document upload (DOCX, XLSX, PPTX)


coolhandle01:
## Summary

Office Open XML formats (DOCX, XLSX, PPTX) are ZIP archives containing XML files. When a server extracts and parses these archives (content indexing, preview generation, mail merge, data import), XXE can trigger inside the inner XML documents.

## Relation to existing work

Companion to the SVG upload issue and #49 (generic XXE probe). OOXML is the third major file-upload XXE surface after SVG.

## Attack surface

Endpoints accepting `application/vnd.openxmlformats-officedocument.*` or multipart uploads with `.docx`/`.xlsx`/`.pptx` extensions. Common in: HR/recruitment platforms (CV upload), document management, mail-merge features, spreadsheet import, reporting tools.

## Blocker

Authenticated file upload flows require the pipeline to have a valid session cookie or Bearer token. This is blocked pending session/auth management work. The tool should accept optional auth headers as a parameter so it can be used once auth is implemented.

### Payload construction

A minimal DOCX is a ZIP containing at minimum:

- `[Content_Types].xml`
- `word/document.xml` - embed XXE SYSTEM entity here
- `_rels/.rels`

Python's `zipfile` (stdlib) can construct this in memory without writing to disk. No third-party deps needed.

```python
import io, zipfile

def _make_docx_xxe(uri: str) -> bytes:
    buf = io.BytesIO()
    with zipfile.ZipFile(buf, "w") as zf:
        zf.writestr("[Content_Types].xml", CONTENT_TYPES_XML)
        zf.writestr("_rels/.rels", RELS_XML)
        zf.writestr("word/document.xml", _docx_body_xml(uri))
    return buf.getvalue()
```

### Detection

POST multipart with `filename="probe.docx"`. Match `/etc/passwd` or `win.ini` markers in response -> CRITICAL.

## Acceptance criteria

- [ ] Construct a valid minimal DOCX ZIP in memory with XXE payload in `word/document.xml`
- [ ] Do the same for XLSX (`xl/workbook.xml`) and PPTX (`ppt/presentation.xml`)
- [ ] POST multipart to candidate upload endpoints, match sentinel markers -> CRITICAL
- [ ] Accept optional `auth_headers` parameter for use once session management is implemented
- [ ] Unit tests mock all HTTP and use `io.BytesIO` - no disk I/O
- [ ] `bandit`, `ruff`, `mypy` all pass

Issue #74, opened 2026-05-15
SAML vulnerability testing: XXE via SAMLResponse and Signature Wrapping (XSW)


coolhandle01:
## Summary

SAML is an XML-based SSO protocol widely used across HackerOne programmes (enterprise SaaS, IdP integrations, federated login). Two distinct attack classes apply:

1. **XXE via SAML** - If a service provider base64-decodes and parses a `SAMLResponse` POST body without disabling external entity expansion, an attacker-controlled SAMLResponse can trigger an in-band file read (e.g. `/etc/passwd`) or OOB DNS callback. This is distinct from the generic XXE probe (#49) because the payload is form-encoded (`SAMLResponse=<base64>`), not sent as `application/xml`.

2. **SAML Signature Wrapping (XSW)** - The attacker takes a valid, signed SAMLResponse, clones the signed `<Assertion>` with a malicious identity, and wraps/reorders it so the signature validation covers the legitimate node while the application reads the attacker-controlled node. Eight canonical XSW variants (XSW1-XSW8) cover different wrapping positions in the SOAP envelope and XML document. A successful XSW lets an attacker authenticate as any user, including admins.

## Relation to existing work

Closes or extends #68 (OAuth 2.0 and SSO vulnerability testing). SAML is the XML leg of SSO; OAuth/OIDC is the JSON leg.

### Tool location

`tools/pentest/saml.py` - two functions, one `@tool` wrapper in the PT agent.

### XXE via SAML (`check_saml_xxe`)

```
POST /saml/acs  (application/x-www-form-urlencoded)
SAMLResponse=<base64(malicious_xml)>
```

- Craft generic and SOAP-envelope XXE payloads (reuse `_generic_xml` / `_soap_xml` from `tools/pentest/xxe.py`)
- base64-encode and POST to candidate ACS endpoints (discovered via recon: URLs containing `/saml`, `/sso`, `/acs`, `/login/saml`)
- Match `/etc/passwd` or `win.ini` markers in response body -> CRITICAL
- Match XML parser error markers -> MEDIUM (confirms XML parsing without entity guard)

### XSW (`check_saml_xsw`)

Requires a valid `SAMLResponse` as input (captured from an actual login flow - the agent would need to surface this from recon or accept it as a parameter). Given that:

- Parsing and re-signing XML requires `lxml` + `signxml` (not currently in deps)
- The PT agent does not have a login session to capture a real SAMLResponse
[...]

Issue #68, opened 2026-05-14
feat: OAuth 2.0 and SSO vulnerability testing


coolhandle01:
## Summary

No tool currently tests for OAuth 2.0 or SSO vulnerabilities. OAuth is ubiquitous in modern web apps and a reliable source of high-severity H1 findings - open redirect chains into OAuth token theft, state parameter bypass, and redirect_uri manipulation are consistently rewarded. Maps to OWASP A01 (Broken Access Control) and A07 (Identification and Authentication Failures).

## What to implement

New file: `tools/pentest/oauth.py`

### Discovery

Before testing, identify OAuth endpoints from the recon result:
- Look for `/oauth/authorize`, `/oauth2/authorize`, `/connect/authorize`, `/auth/callback`, `/oauth/token` in discovered URLs
- Look for `client_id=`, `redirect_uri=`, `response_type=`, `state=` in URL parameters
- Check for OIDC discovery document at `/.well-known/openid-configuration`

#### 1. `redirect_uri` validation weakness

- Substitute a known-bad redirect URI and check whether the server rejects it or issues an authorisation code/token to the attacker URI:
  - Different subdomain: `https://evil.example.com/callback`
  - Path traversal: `https://legit.example.com/../../evil`
  - Param injection: `https://legit.example.com/callback?next=https://evil.com`
  - Open redirect chaining: if the app has a known open redirect, use it as the redirect_uri
- A 302 to the attacker URI with `code=` or `token=` in the location is a CRITICAL finding

#### 2. Missing or reused `state` parameter (CSRF on OAuth flow)

- Initiate the OAuth flow without a `state` parameter - if the server proceeds without rejecting the request, CSRF on the OAuth flow is possible
- Replay a previously observed `state` value - if accepted, the state is not single-use

#### 3. Authorisation code reuse

- Capture an authorisation code from a completed flow and attempt to exchange it a second time at the token endpoint. If the server issues a new token, codes are not invalidated after use.

#### 4. Token leakage via Referer

- If the access token or authorisation code appears in the URL (implicit flow or code in query string), check whether the app loads any third-party resources that would receive the token via the `Referer` header.

#### 5. PKCE bypass (public clients)

- For endpoints that advertise PKCE support, attempt the token exchange without supplying a `code_verifier`. If the server issues a token anyway, PKCE is not enforced.

#### 6. SSO login CSRF / account linking abuse

[...]

PR #160, opened 2026-05-25
feat(security): Pydantic-typed agent tool surface - LLM-safe inputs across every wrapper

### LINKED ISSUE #156: refactor: post-#150 typed-primitive gaps + cloud scope gate + programme-handle dedupe

coolhandle01:
## Why

Before opening cybersquad's tool surface to MCP servers (#144) - or any caller beyond the in-process CrewAI runtime - every wrapper's input contract has to be safe against the LLM mis-calling it. An LLM that picks an out-of-scope target, hands us a malformed URL, or smuggles a case-variant duplicate header is the input-side prompt-injection class ([OWASP LLM01:2025](https://genai.owasp.org/llmrisk/llm01-prompt-injection/)); the LLM-facing surface needs to reject those at the boundary, not in the wrapper body and not at the HTTP layer.

The shape that closes the gap end-to-end is **Pydantic-typed `args_schema`s with validators at the boundary** - Alexis King's [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) applied to LLM-facing surfaces. Pydantic v2 runs every validator during `args_schema.model_validate(...)`; CrewAI runs that before any tool body sees input; so the typed field IS the contract.

This PR makes every wrapper carry one.

### Scope guard as typed `args_schema` fields

Four typed aliases in `tools/recon/scope.py` carry the scope filter as a Pydantic `AfterValidator`. The LLM picks targets and the type rejects out-of-scope picks before the wrapper body sees anything:

| Alias | Inner | Semantic |
|---|---|---|
| `TargetHostnames` | `list[Hostname]` | Silent filter - mixed candidate lists, in-scope survive |
| `TargetEndpoints` | `list[Endpoint]` | Same filter, host extracted from `.url` |
| `TargetHostname` | `Hostname` | Loud reject - a single target is the LLM committing |
| `TargetEndpoint` | `Endpoint` | Same loud-reject |

Every `args_schema` across `osint_analyst/`, `penetration_tester/cloud/`, `penetration_tester/probes/` declares one - ~40 fields. The field type IS the scope guarantee. The validator reads `current_programme()` from the run-dir snapshot; the LLM never sees, names, or threads the programme handle.

### URL validation delegates to upstream

`models.primitives.HttpUrl` validates via [`pydantic.HttpUrl`](https://docs.pydantic.dev/2.12/api/networks/#pydantic.networks.HttpUrl) (canonical RFC-3986 parser); the host component then runs through the `Hostname` validator for RFC 1123 strictness (defense-in-depth - Pydantic's parser is RFC-3986-permissive on the host, which is broader than the bounty surface should be probing). Runtime type stays `str` so existing consumers don't pay an audit cascade.

### Outbound headers dedupe by default

`tools/http.py`'s default merge target is [`requests.structures.CaseInsensitiveDict`](https://github.com/psf/requests/blob/main/src/requests/structures.py). [RFC 9110 section 5.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1) makes header field names case-insensitive; the safe default never accidentally ships duplicates from a `{"User-Agent": ..., "user-agent": ...}` mistake. Header-mutation probes that genuinely need literal duplicates (request smuggling, CRLF, host-header confusion, HPP) work around the safe default deliberately - auditable as the explicit pentester action in the probe source.

### Test surface reorganised

`tests/conftest.py` shrunk 553 -> 62 lines; fixtures grouped by concern under `tests/fixtures/*.py` (domains / programme / recon / findings / responses / tools) and wired in via `pytest_plugins`. A new `invoke_tool` fixture mirrors CrewAI's call path (`args_schema.model_validate(...).model_dump()` -> body), so tests that exercise the typed scope guard exercise the same validator the runtime does.

## Further reading

The README's "Scope and safety" section now describes the actual landed mechanism (it previously referenced the prior wrapper-side machinery).

- [OWASP Top 10 for LLM Applications - LLM01:2025 Prompt Injection](https://genai.owasp.org/llmrisk/llm01-prompt-injection/) - the threat model the typed surface defends against
- [Alexis King, "Parse, don't validate" (2019)](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) - the canonical reference for boundary-typed validation
- [Pydantic v2 docs](https://docs.pydantic.dev/2.12/concepts/models/) - the runtime that makes "parse don't validate" cheap in Python
- [RFC 9110 section 5.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1) - HTTP header field name case-insensitivity (the property `CaseInsensitiveDict` enforces)

Skills updated alongside: `cybersquad-tool` carries the typed-alias canonical example and the new scope-guard section; `cybersquad-pentest-tool` cross-links; `cybersquad-test-fixtures` describes the new fixture-module layout.

## Tests

`1618 unit tests pass; combined line + branch coverage 93.43%; ruff / ruff format / mypy / pylint (9.90/10) / bandit clean.`

Closes #156.


coolhandle01:
> Storage (s3, azure_storage) keeps recon_path: str. 

I would like to push back on that, all pentest tools should use the scope_checkguardfilter and args_schema, even the cloud ones, which will either be in scope or ruled out by the programme scope in H1 and enshrined in the Programme model somewhere. .. maybe this is word soup, alas:

- PM: choose program, transcribe scope
- OSINT Analyst: discovery of surface/inventory of scope/assets
- VR: analyse scope and plan attack(s)
- PT: planned attacks on scope assets
- VR attack analysis/remediation 
- TA..
- DC..

coolhandle01:
Fair pushback. You're right that the "storage exception" was a smell — and digging into _why_ storage was exceptional, the answer was that `tools/cloud/aws.py` and `tools/cloud/azure.py` were doing **bucket-name fuzzing** (`_bucket_candidates` glued the programme handle onto 10 suffixes — `myapp`, `myapp-assets`, `myapp-backup`, ... — and probed `*.s3.amazonaws.com` for each guess). That's the actual problem: probing third-party tenants you guessed might be the programme's is not legit bounty methodology.

Pushed 48cdc5c which:

- **Removes** the suffix-fuzzing in both `aws.py` and `azure.py` entirely. We only probe assets OSINT actually discovered in `recon.subdomains`.
- **S3 Bucket Check** → `list[Hostname]` + `scope_filter=("hostnames", filter_in_scope)`. Same shape as databases / panels / etc.
- **Azure** splits into two tools:
  - `Azure Blob Container Check` → `list[Hostname]` + scope_filter. Canonical container-name list (`public`, `assets`, `static`, ...) is probed against each supplied hostname (equivalent to the canonical `/admin` paths in `Admin Panels Check` — that's targeted probing of known patterns on a discovered host, not customer-name guessing).
  - `Azure SAS Token Check` → `list[Endpoint]` + scope_filter. Static URL inspection only; no HTTP fires.

Tool count 22 → 23. The cloud package docstring drops its FIXME entirely — no third-party-infra exception anymore. Aligns with your pipeline framing end-to-end:

> PM transcribes scope → OSINT inventories the surface → PT attacks the inventoried surface

1614 unit tests pass; 93.34% coverage holds.

---

tests/test_cloud.py
```diff
@@ -131,82 +98,51 @@ def recording_get(url, **kwargs):
...
 
-        assert any("myapp-assets" in u for u in seen_urls)
+        for hostname in hostnames:
+            assert any(hostname in u for u in seen_urls)
 
-    def test_network_exception_is_swallowed(self, minimal_recon):
+    def test_network_exception_is_swallowed(self):
         with patch("requests.get", side_effect=Exception("timeout")):
-            results = check_s3_buckets(minimal_recon)
+            results = check_s3_buckets(["myapp.s3.amazonaws.com"])
```
coolhandle01: Could these be `f"{target_apex}.s3.amazonaws.com"`, just for sake of clarity of intent?

tests/test_cloud.py
```diff
@@ -216,15 +152,59 @@ def recording_get(url, **kwargs):
...
+            results = check_azure_sas_tokens([endpoint])
+
+        sas_findings = [r for r in results if "SAS Token" in r.title]
+        assert len(sas_findings) == 1
+        assert sas_findings[0].severity_hint == Severity.HIGH
+        # Static URL inspection: never fires HTTP.
+        mget.assert_not_called()
+
+    def test_clean_urls_produce_no_findings(self):
+        endpoint = Endpoint(url="https://app.example.com/files/doc.pdf?download=1", status_code=200)
```
coolhandle01: target_url?

tools/cloud/aws.py
```diff
@@ -4,57 +4,36 @@
 
 import logging
 
-from models import RawFinding, ReconResult, Severity
```
coolhandle01, line 7: The ReconResult here in this file makes me think of #83 - maybe aws/azure/gcp etc is a framework detection, (#68 #74 #76 etc)

squad/penetration_tester/cloud/storage.py
```diff
@@ -1,65 +1,121 @@
...
 
-See the package-level scope-of-target FIXME in ``cloud/__init__.py``
-(tracked in #156).
+Each wrapper takes a typed target the agent picks from recon (the
+``*.s3.*.amazonaws.com`` / ``*.blob.core.windows.net`` hostnames the
+OSINT Analyst discovered, or the endpoints whose URLs contain SAS
+tokens) and a wrapper-level ``scope_filter`` drops anything outside
+the selected programme's structured scope before the probe fires.
+
+No bucket-name or account-name guessing: every probed asset is one
```
coolhandle01: If doesn't exist then don't document it

squad/penetration_tester/cloud/storage.py
```diff
@@ -1,65 +1,121 @@
...
 
 from pydantic import BaseModel, Field
 
-from models import RawFinding
+from models import Endpoint, Hostname, RawFinding
 from squad import cyber_tool
-from squad.penetration_tester._decorator import _recon_from_path
-from tools.cloud import check_azure_storage, check_s3_buckets
+from squad.penetration_tester._decorator import _parse_endpoints
+from tools.cloud import check_azure_blob_containers, check_azure_sas_tokens, check_s3_buckets
```
coolhandle01: I think I'd like to see `from tools.cloud.azure import check_szure_ ...`, ,`from tools.cloud.aws import check_s3_buckets`

squad/penetration_tester/cloud/web_content.py
```diff
@@ -4,10 +4,11 @@
 artefacts) and for admin-panel login pages exposed at predictable
 paths.
 
-Unlike the other cloud / infra wrappers in this package these take
-typed ``list[Endpoint]`` (the agent picks which endpoints to ask
-about) rather than a recon path - they are HTTP path probes, not
-host-derived service checks.
+These have always taken typed ``list[Endpoint]`` (the agent picks
```
coolhandle01: Make it timeless etc

squad/penetration_tester/cloud/web_content.py
```diff
@@ -55,21 +61,25 @@ class _AdminPanelsArgs(BaseModel):
...
+)
 def admin_panels_tool(endpoints: list[Endpoint]) -> list[RawFinding]:
     """
-    Probe common admin panel paths: /admin, /wp-admin, /phpmyadmin, /adminer,
-    /manager/html, /_admin. Run broadly on all live endpoints.
-
-    endpoints: list of endpoint objects. The tool deduplicates by origin.
-      Example: [{"url": "https://example.com/", "status_code": 200}]
-
+    Probe common admin panel paths: /admin, /wp-admin, /phpmyadmin,
```
coolhandle01, line 68: This is a bit hazardous (glad this PR brought my attention to it). This is a classic case of StrEnum. Gutted: you get to break ground in #83 .. just, be unopinionated/KISS for here and now, minimal shift.. just make the Panel "brand"(sic "tech" aka "suspected-tech" aka "technology") the "attack type" like a pentest_tool?

tests/test_cloud.py
```diff
@@ -31,25 +34,15 @@
...
 
 
 class TestCheckS3Buckets:
-    def test_detects_publicly_listable_bucket(self, minimal_recon):
+    """The agent picks S3 hostnames OSINT actually surfaced in recon;
+    no bucket-name guessing - we only probe assets the programme has
+    exposed."""
+
+    def test_detects_publicly_listable_bucket(self):
         listing_xml = (
```
coolhandle01: Haza I Haz found (*knights who say ni voice*)... A fixcharr!! 

If I'm learning anything then there should be fixtures to back these azure/aws checks

tests/test_cloud.py
```diff
@@ -58,66 +51,44 @@ def test_detects_publicly_listable_bucket(self, minimal_recon):
         mock_resp.text = listing_xml
 
         with patch("requests.get", return_value=mock_resp):
-            results = check_s3_buckets(minimal_recon)
+            results = check_s3_buckets(["myapp-assets.s3.amazonaws.com"])
```
coolhandle01: Cries in target_url and a desire for "composite" fixtures people always find and use when they need to

tools/cloud/azure.py
```diff
@@ -21,70 +21,80 @@
...
+    Scan each supplied endpoint URL for embedded Azure SAS-token query
+    parameters (``sv`` / ``se`` / ``sig`` / ``sr`` / ``sp``). Static
+    URL inspection - no HTTP requests. SAS tokens embedded in URLs are
+    logged by intermediate proxies and browser history; surfacing them
+    is the finding.
+    """
+    findings: list[RawFinding] = []
+
+    for ep in endpoints:
+        if _SAS_RE.search(ep.url):
```
coolhandle01, line 96: Do we have appropriate model ceremony for changes here?

tools/cloud/services.py
```diff
@@ -74,17 +74,6 @@ def _unique_origins(endpoints: list[Endpoint]) -> list[str]:
     return result
```
coolhandle01, line 76: Sounds like useful query on ReconResult (that would transpose well to an amass-centric future)  🤔

tools/recon/scope.py
coolhandle01: Looks like it should be defined somewhere more low-level, this `extract_domain` .. should this be part of the HttpUrl thing that I think we were going to replace with the Pydantic one but maybe it needed some work?

coolhandle01, line 58: I think this and it's new sibling are more low-level or "core" than "tools/recon/scope.py"

Is Scope a model? .. this being "up here" feels odd when scope_filter is "down there".. what's the code telling us, oh python master? 🥋

coolhandle01 (commented):
Lots of spots to polish and thus opportunities to make sure fixtures and parameterize pytest features are used most optimally and tests are trustworthy. When these issues are addressed, please recommend updates or improvements to the contributor skills in the repo

coolhandle01:
Pushed 165b9eb with the mechanical fixes:

- **Fixture factories** in `conftest.py`: `target_sld` (the SLD `example` from `example.com`, since bucket / account-name labels can't contain the apex's dot), `make_s3_hostname` / `s3_hostname`, `make_azure_blob_hostname` / `azure_blob_hostname`, `azure_sas_endpoint`. Mirrors the `make_html_page` pattern - single value for the common case, factory when a test needs variants.
- **`test_cloud.py`** uses the fixtures end-to-end: every `myapp.s3.*` and `app.example.com` literal flipped.
- **Timeless docstrings** in `storage.py`, `web_content.py`, `aws.py`, `azure.py` - "no fuzzing" / "have always taken" history-prose dropped (principle stays codified in the `cybersquad-tool` skill anti-pattern where new tool authors see it).
- **Direct submodule imports** in `storage.py` (`from tools.cloud.aws import ...` rather than via the package `__init__`).

## Architectural smells

Triaging your comments + a couple I'd add. Honest read on each:

### Cloud provider as framework-detection (`aws.py:7`)
**Agree, files-as-sub-issue.** AWS / Azure / GCP detection (from response headers, hostname patterns, ASN lookups) is the same shape as Django / Rails / Spring detection in #83. Belongs in `tools/pentest/framework/cloud.py` once #83 grows beyond one file. Worth a comment on #83 or a sub-issue. **Defer.**

### Admin-panels brand as StrEnum (`web_content.py:79`)
**Agree, defer to #83.** `check_admin_panels` has a hardcoded `_ADMIN_PATHS` list with no agent input on brand. The pentest-tool pattern (`AdminPanelBrand` StrEnum + `check_fn` filter via `@pentest_tool`) would let the agent target specific brands once they're detected by the framework-detection work. KISS for here; this PR keeps it as-is. **Defer.**

### `_unique_origins` as ReconResult query (`services.py:76`)
**Agree, files-as-sub-issue.** Extracting unique origins from `recon.endpoints` is a reusable query - currently lives as a `_`-prefixed helper in `services.py`. Belongs as either `ReconResult.unique_origins` (method) or a typed slicer `Recon Unique Origins` (next to the existing `Recon Endpoints` / `Recon Open Ports` / `Recon Subdomains`). Latter is more agent-facing. **Worth a sub-issue.**

### `extract_domain` on `HttpUrl` primitive (`scope.py:18`)
**Agree, already half-tracked.** `models/primitives.py` already has a FIXME about moving to Pydantic's built-in `HttpUrl` (which exposes `.host` natively and would make `extract_domain` unnecessary). Right move is to reference that FIXME from `scope.py` so they lift together. **Small follow-up in this PR or the next - say the word.**

### `tools/recon/scope.py` placement (`scope.py:24`)
**Agree, real smell.** Scope is now cross-cutting (OSINT recon-time filter + every PT cloud wrapper + the `@cyber_tool(scope_filter=...)` decorator). Living under `tools/recon/` is historical (recon was the first consumer). Two moves:
- `tools/scope.py` top-level - small diff, just import shifts in maybe 5 files.
- `models/scope.py` if we promote `Scope` to a first-class model (`ScopeItem` already exists in `models/h1.py`). Bigger structural call.

I'd do the `tools/scope.py` move in **this PR** if you want - low-risk, alignment matches its actual reach. The Scope-as-model question is a separate design call. **Say the word.**

### Model ceremony at `azure.py:95`
**Unclear what you mean here.** Line 95 is the `RawFinding` construction for SAS-token findings. The existing typed `RawFinding` is the model ceremony as I read it. Got something specific in mind - a `SasTokenFinding` subclass? a parsed-SAS-params model on `RawFinding.evidence`? Happy to chase whichever, just need a steer.

### My own additions (honest hunt)

- **`filter_endpoints_in_scope` accepts `Endpoint | Mapping[str, Any]`**. The both-shapes adapter couples the scope helper to CrewAI's call shape. The natural home for the dict-branch is `_apply_scope_filter` in `squad/__init__.py` (next to where `args_schema.model_validate` already normalises) so the scope helper signature stays clean. Small refactor; **could do in this PR**.
- **`_HOSTNAME_PASSING_*_WRAPPERS` test tables duplicate `MEMBER.tools`**. Three places (`MEMBER.tools`, `_PT_SCHEMAS`, the parametrize tables) to update for one new wrapper. The contract-test in `test_pt_args_schemas.py` already walks the registry; the parametrize tables could derive from it the same way. **Sub-issue if you'd like.**
- **Cloud wrappers don't normalise hostnames**. OSINT's `_normalise_and_filter_hostnames` strips / lowercases before scope-filtering; cloud wrappers go straight to `filter_in_scope`. If the agent ever hands us `" API.Example.Com\n"`, OSINT handles it; cloud rejects it. Either lift the normaliser to a shared helper (covered by both) or document why cloud doesn't need it. **Small fix in this PR or the next.**

---

**Concrete next steps if you want to land them in this PR**:
1. Move `tools/recon/scope.py` → `tools/scope.py` (cross-cutting placement).
2. Move the both-shapes branch out of `filter_endpoints_in_scope` into `_apply_scope_filter`.
3. Lift the hostname normaliser to a shared helper used by both OSINT and cloud wrappers.

Or carry any subset to follow-up issues. Your call - happy to do all three, none, or a subset.

---

main.py
```diff
@@ -69,6 +69,11 @@ def check_env() -> None:
         sys.exit(1)
 
 
+# ``Any`` here is deliberate: CrewAI's ``Crew`` is wrapped through several
+# decorators by the time this CLI handler sees it, and ``Crew | object``
+# adds no useful constraint at the boundary. ruff: noqa ANN401 - see
+# https://docs.astral.sh/ruff/rules/any-type/ for the rule's rationale
```
coolhandle01: Booo this comment is long and I like self-documenting code, plus, is the performance benefit really _that_ significant?

coolhandle01 (commented):
I think good progress is being made in decluttering the code, but I think it's worth looking at the remaining ones and discussing what the code tells us it is missing. I feel like the overall wind down of the typesafety drive is there but was rushed at the end and now needs crystalising more concretely.

coolhandle01:
Fresh-eyes review pass; pushed 3e3caf0 addressing the three open threads.

**Landed:**

1. **`main.py:75`** - 5-line block comment for one `# noqa: ANN401` collapsed to single line on the directive, matching CONTRIBUTING.md's one-line `why` grammar.
2. **`tests/test_cloud.py`** - the new S3 / Azure / `TestGranularPanels` rewrites were hand-rolling `MagicMock().status_code = ...` when `conftest.make_response` already exists for exactly this (the cybersquad-test-fixtures skill flags it as the canonical pattern). Threaded `make_response` through; tightened 19 call sites. Pre-existing `MagicMock` patterns in databases / sensitive-files / admin-panels classes left alone per minimal-diff.
3. **`web_content.py` "make it timeless"** - already addressed in 165b9eb; the current docstrings carry no historical phrasing. Resolving on next pass.
4. **Skill catalogue update** - `cybersquad-test-fixtures` SKILL.md catalogue gains `target_sld`, `make_s3_hostname` / `s3_hostname`, `make_azure_blob_hostname` / `azure_blob_hostname`, `azure_sas_endpoint`. The four cloud-storage fixtures this PR added should be discoverable from the skill, not just from conftest.

`1619 unit tests pass; coverage 93.38%; ruff / mypy / pylint / bandit clean.`

**Deferred — surfacing for your call:**

The "rushed wind-down" you flagged shows up in a couple of structural places I read but did not act on:

- **Reverted decoration-time guard (c7c2dc6 → 4474562, no reason in the revert commit).** The guard would have raised `TypeError` at `@cyber_tool` decoration time on unsupported scope-target shapes (e.g. `dict[Hostname, int]`, nested models). Today auto-detection silently does nothing on those shapes - a future tool author who picks an unsupported shape gets no signal. If the revert was tactical (e.g. it broke a real existing shape we want to keep), worth a FIXME with the rationale; if it was just "scope creep for this PR", worth a sub-issue.
- **`filter_endpoints_in_scope` accepts `Endpoint | Mapping[str, Any]`.** The both-shapes branch was previously flagged as belonging in `_apply_scope_filter` so the scope helper signature stays clean. Three places handle the same ambiguity (`_apply_scope_filter` → `filter_endpoints_in_scope` → `_parse_endpoints`); the seams could be one boundary.
- **`tools/recon/scope.py` placement.** Now cross-cutting (recon + every cloud wrapper + `@cyber_tool` introspection). The `tools/recon/` location is historical.
- **`extract_domain`.** Belongs on `HttpUrl` once the Pydantic `HttpUrl` migration lands (already FIXME'd in `primitives.py`).

None of these are blockers for landing #160 IMO - they are follow-up sub-issues. Happy to file them or carry any into this PR, whichever you prefer.

---

main.py
coolhandle01: Hmmmmmmmmm fine for now I guess. I tend to take this sort of linter technicality situation as a hint from the code at some existential wider misunderstanding - without having looked at the docs, I now feel like "Any being fine at the CLI boundary according to CrewAI" is probably the documentation of where CrewAI had t compromise on some principle to make something work.

I will ponder it elsewhere.

coolhandle01:
Put the slop to bed — pushed three commits:

**b1382ee — pydantic-native scope guard, everywhere.** Four typed aliases in `tools/recon/scope.py`:

```python
InScopeHostnames = Annotated[list[Hostname], AfterValidator(...)]   # filter list
InScopeEndpoints = Annotated[list[Endpoint], AfterValidator(...)]   # filter list
InScopeHostname  = Annotated[Hostname,       AfterValidator(...)]   # reject single
InScopeEndpoint  = Annotated[Endpoint,       AfterValidator(...)]   # reject single
```

CrewAI's normal `args_schema.model_validate(...)` runs the filter — no decoration-time introspection, no `_apply_scope_filter`, no `_scope_filter_for_field`, no both-shapes adapter. `@cyber_tool` shrinks from ~150 lines to ~10 (just attaches the args_schema). All 40-ish args_schema fields across `osint_analyst/`, `penetration_tester/cloud/`, `penetration_tester/probes/` switched. `extract_domain` deleted; three callers use stdlib `urlparse(...).hostname` (the new one-line `host_of` helper). `filter_endpoints_in_scope` gone — its both-shapes branch existed because the auto-detect path saw dicts; with Pydantic owning validation, only typed instances reach the filter.

**5451341 — HttpUrl delegates to `pydantic.HttpUrl`.** URL parsing is now [Pydantic's canonical validator](https://docs.pydantic.dev/2.12/api/networks/#pydantic.networks.HttpUrl). Runtime type stays `str` so the 50+ `ep.url.startswith(...)` / `.lower()` / `urlparse(ep.url)` consumers under `tools/pentest/*` don't have to migrate in lockstep — that audit is tracked separately as #163. The thin layer around upstream does only what upstream doesn't: reject `https:///path` (empty authority — Pydantic treats `path` as the host), and apply the RFC 1123 `_validate_hostname` to the host component (Pydantic's URL parser is RFC-3986-permissive on leading-hyphen labels, which is broader than the bounty surface should be probing).

**a5029e3 — `requests.structures.CaseInsensitiveDict` in `tools/http.py`.** Free security from the [right primitive](https://github.com/psf/requests/blob/main/src/requests/structures.py): RFC 9110 §5.1 makes header field names case-insensitive, so the default merge dedupes by lower-cased key. A plain `dict` accepting `{"User-Agent": "ours", "user-agent": "theirs"}` ships duplicates silently; this default ships one well-defined entry. The docstring comments document both the safe default *and* the override pattern — a header-mutation probe (request smuggling, CRLF, duplicate-Host) that needs literal duplicates works around `CaseInsensitiveDict` deliberately, which makes the attack auditable as such in the probe's source. New regression test pins the dedupe contract.

**Skills updated alongside** — `cybersquad-tool` canonical example switches to `InScopeHostnames` + new "Scope guards" section around the four aliases; `cybersquad-pentest-tool` SSRF example switches to `InScopeEndpoints` and cross-links.

1618 unit tests pass; coverage 93.43%; ruff / format / mypy / pylint / bandit clean. No reverts in this lot; no sub-issues for the work itself.

---

PR #113, opened 2026-05-17
feat: Textual TUI, headless mode, and runtime state module

coolhandle01:
## Summary

- **`runtime.py`** — new module: `run_id`, `programme_handle`, `run_dir()`, `programme_cache_path()`
- **`tools/tui/__init__.py`** — new generic `CrewAIPipelineTUI(App)` base class; crew, task_map, record_prefix, verbose, and dry_run are constructor-injected; `_make_task_callback` / `_make_step_callback` are instance methods
- **`tui.py`** — 19-line `CybersquadTUI(CrewAIPipelineTUI)` wrapper; sets `CSS_PATH`, builds crew, passes `record_prefix="cybersquad"`
- **`tui.tcss`** — Textual CSS, separate file
- **`squad/__init__.py`** — `SquadMember.phase` → `.task`
- All 6 squad member files — `phase="..."` → `task="..."`
- **`crew.py`** — `squad_phases()` → `crew_tasks()`
- **`main.py`** — `--headless` flag; `--dry-run` and `--verbose` apply to TUI mode too
- **`pyproject.toml`** — `textual>=0.80.0` dep; `tools.tui` package; TUI files excluded from coverage

Replaces #109 (wrong branch name).

## Test plan

- [x] `ruff check .` — clean
- [x] `ruff format --check .` — clean
- [x] `mypy . --ignore-missing-imports` — clean
- [x] `pytest -m unit --cov` — 735 passed, 90.55% coverage
- [x] `bandit -c pyproject.toml -r . -q` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01B6R3CrztXjyZwMsidwxiWS)_

Issue #128, opened 2026-05-19
spike - go with the flow


coolhandle01:
when we land #114 , we'll have everything we need to evolve the _Pipeline_ into a _Flow_ that lets the team work more organically.

Consider the following example produced by Gemini:

```python
from crewai.flow.flow import Flow, listen, start

class SupportFlow(Flow):
    @start()
    def initial_event(self):
        # This acts as your entry point event
        return {"status": "new_ticket", "priority": "high"}

    @listen(initial_event)
    def route_task(self, event_data):
        if event_data["priority"] == "high":
            # Direct to a specific crew or agent logic
            print("Escalating critical event...")
```

With listen, and start, feedback loops can be used give the squad more atonomy. Maybe the DC and TA can co-ordinate on responses to comments from the triage team on H1 tickets, or the DC can notify PM on the status of the report as it changes so that the PM can keep track of token-> bounty RoI KPIs.

coolhandle01:
_Love note from past me, written during #140 review._

When this lands and the Pipeline becomes a Flow, `tasks.py` will almost certainly stop being one file and start being a module/package - per-agent task files, per-flow listener files, something like that.

The `cybersquad-task` skill matcher (`.claude/hooks/load-skill.sh`) currently fires only on `*/tasks.py`. The day the file goes away or moves, the skill goes dark for those files without anyone noticing. That's classic **skill drift**: the contract the skill encodes is still true, but the skill no longer loads where it's needed, so AI contributors quietly stop applying it.

When you do the Flow migration, update the matcher in the same PR. Candidates:
- `*/tasks/*.py` if it becomes a package
- `*/flows/*.py` for the new shape
- Whatever you choose, add a regression check in the PR description: pipe a sample path through `.claude/hooks/load-skill.sh` and confirm `cybersquad-task` is emitted.

Lots of love,
past me 💙

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #166, opened 2026-05-26
feat: Textual TUI, headless mode, and runtime state module

coolhandle01:
## Summary

Textual TUI surfacing live agent activity during a cybersquad run, a `--headless` flag for sessions without a terminal (CI, scripts), and the `crew_tasks()` role-to-task mapping the TUI consumes.

- **`tools/tui/__init__.py`** - generic `CrewAIPipelineTUI(App)` base class; `crew`, `task_map`, `record_prefix`, `verbose`, `dry_run` are constructor-injected; `_make_task_callback` / `_make_step_callback` are instance methods.
- **`tools/tui/_helpers.py`** - pure helpers extracted for unit-test coverage independently of the App lifecycle.
- **`tui.py`** - 19-line `CybersquadTUI(CrewAIPipelineTUI)` wrapper; sets `CSS_PATH`, builds crew, passes `record_prefix="cybersquad"`.
- **`tui.tcss`** - Textual CSS, separate file.
- **`squad/__init__.py`** - `SquadMember` gains `task: str = ""` alongside the typed `tools: list[SquadTool]` (#160's contract).
- All 6 squad member `__init__.py`s - set their `task=` value.
- **`crew.py`** - `crew_tasks()` returns the role-to-task mapping. Also fixes a pre-existing `from crewai.memory.memory import Memory` that did not resolve in `crewai>=1.14` (1.14.5); `Memory` is exposed at the top level. The new `test_tui_helpers.py` is the first thing that imports `crew` and would otherwise have surfaced it.
- **`main.py`** - `--headless` flag; `--dry-run` / `--verbose` apply to TUI mode too.
- **`pyproject.toml`** - `textual>=0.80.0` dep; `tools.tui` package declared; TUI files excluded from coverage.
- **`tests/test_tui_helpers.py`** - 18 tests covering the pure helpers and the wrapper wiring.

## Provenance

Replaces #113. Same work, on a fresh branch off current main, with the original TUI commits preserved by SHA in this branch's second-parent ancestry (`git log d6cd212` walks them: `09d6c9f`, `8013f5e`, `b369c0f`, `c886d3a`, `4d2e348`, `89f6e8a`, `500d946`, `1eb1cff`).

## Test plan

- [x] `pytest -m unit --cov` - 1636 passed, combined line+branch coverage 93.65%
- [x] `ruff check .` - clean
- [x] `pylint .` - 9.90/10

---
_Generated by [Claude Code](https://claude.ai/code/session_01T89UNyww6VkAsyWPVHe5Vr)_


coolhandle01:
You're not mistaken — the `task` field is the noisiest part of this PR. Two things.

**1. Naming.** The field was `phase` in the original TUI commits and was renamed to `task` in 8013f5e (`refactor(tui): extract generic CrewAIPipelineTUI; rename phase→task`). The rename was a regression: `task` collides with two existing concepts on the same dataclass —

- CrewAI's `Task` — the seven pipeline tasks built in `tasks.py`.
- The task-subdir names already passed to `build_task("submit", ...)` — `select/`, `recon/`, `research/`, `pentest/`, `triage/`, `write/`, `submit/`.

`SquadMember.task = "Disclosure"` is neither. It's a **phase label** for the TUI sidebar — `phase` was the right name and was self-consistent: the TUI helper still uses `phase` internally as a variable (see `task_phase_layout` computing `phase = task_map.get(name, name)`), the rename only landed on the dataclass side and broke the symmetry. Cheapest fix: revert to `SquadMember.phase` and `crew_phases()`. The CSS class `.phase-heading` is already on team `phase`.

**2. Required only for VR.** `task_phase_layout` already falls back to the task-name (i.e. the role) when a member has no phase set, so for the five single-task members the field is decorative — without it the sidebar would read "Disclosure Coordinator", "Penetration Tester", etc. instead of "Disclosure", "Penetration Testing". The field is *strictly* required only for the Vulnerability Researcher, where two pipeline tasks share an agent and would otherwise emit two identical "Vulnerability Researcher" rows under one heading. (Worth noting that even with `task="Vulnerability Research"` set, the current dedup means both VR rows still read identically; you might want the per-task subdir name surfaced there separately, but that's a follow-up.)

Reasonable to keep phase strings on every member for sidebar aesthetics — just be explicit that `task="Disclosure"` on the Disclosure Coordinator is doing a small bit of work, not none, and that the dataclass coupling is a choice. A cleaner home for the role→phase mapping might be `tasks.py` next to where the CrewAI Tasks themselves are built; `SquadMember` could then go back to identity + tools + prose location.

Rest of the diff looks tight: pure-helpers extraction is the right shape, branch coverage on the helpers is solid, and `from crewai import Memory` is the documented public path on 1.14.5+.

---

coolhandle01:
Is this useful/fun https://pypi.org/project/scrolltext/#description ?

tui.py
```diff
@@ -0,0 +1,32 @@
...
+Launch with: python main.py  (default) or python main.py --headless to skip the TUI.
+"""
+
+from __future__ import annotations
+
+from mcp_servers import ProvisionedMCPTools
+from tools.tui import CrewAIPipelineTUI
+
+
+class CybersquadTUI(CrewAIPipelineTUI):
```
coolhandle01: I've got a comment here in the CrewAIPipelineTUI that this line proves is silly :)

I think the CrewAIPipelineTUI should own tui.tcss, and cybersquad's TUI should be able to override it

tui.py
```diff
@@ -0,0 +1,32 @@
...
+    CSS_PATH = "tui.tcss"
+
+    def __init__(
+        self,
+        verbose: bool = False,
+        dry_run: bool = False,
+        mcp_tools: ProvisionedMCPTools | None = None,
```
coolhandle01: I feel like relying on build_crew here shows how CrewUI isn't really reusable. CybersquadTUI shouldn't creally care about MCP tools at this interface.. maybe a out a crew, sure. Hmmm.

main.py
coolhandle01: This main is a bit all over the place.
- Why wouldn't a dry run show metrics? It would be pretty reassuring!
- why not pass dry_run to build_crew (why isn't build_crew doing the MCP init??), then just pass the crew to the TUI?
- all this stuff being imported after dry run is checked, big code smell

tools/tui/_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+
+def task_layout(tasks: list) -> list[tuple[str, str]]:
+    """Build the sidebar entries for a sequential pipeline.
+
+    For each task that has an assigned agent, return a ``(heading, role)``
+    pair in pipeline order. ``heading`` is the task's display name -
+    ``Task.name``, stamped from the member's ``name.md`` - falling back to the
+    agent role when a task carries no name; ``role`` is the agent role shown on
+    the task's status row. Because the heading is per-task rather than
+    per-agent, an agent that runs more than one task (the Vulnerability
```
coolhandle01: Try to keep examples in the tools/TUI/ files not-cybersquad centric but more generic.

coolhandle01 (commented):
Just a few bits with a view to eventually lifting this all into its own package

crew.py
```diff
@@ -138,3 +140,25 @@ def build_crew(
...
+    (``main.py`` and its renderers) gets a finished crew without touching MCP
+    provisioning itself. The MCP subprocesses stay up for the lifetime of the
+    ``with`` block, so that block must enclose ``crew.kickoff()`` /
+    ``App.run()`` - per the ``cybersquad-mcp`` skill (Rule 2, build-time only).
+
+    ``dry_run`` skips MCP startup entirely: a preview never calls a tool, so it
+    must not spawn subprocesses. The yielded crew then carries no MCP-sourced
+    tools (``build_crew``'s ``mcp_tools=None`` path).
+    """
+    if dry_run:
```
coolhandle01: Why can't this be as simple as

```python
mcp_tools = [] if args.dryrun else provisioned_mcp_tools()
yield build_crew(verbose=verbose, mcp_tools=mcp_tools)
```

coolhandle01:
> **Draft bug report for the CrewAI repo** (`crewAIInc/crewAI`) — not about this PR; parked here so it can be lifted/pasted upstream. Documents why this repo pins `crewai==1.14.4`.

---

**Title:** `human_input=True` degraded across releases: result hidden unless `verbose` (1.14.4), then crashes outright on the experimental `AgentExecutor` (1.14.5+)

### Summary

Human-in-the-loop (`Task(human_input=True)`) has regressed over recent releases:

- **1.14.4:** the `💬 Human Feedback Required` panel ("Provide feedback on the Final Result above") is shown, but **the result is never printed** unless the agent/crew is `verbose` — so the operator is asked to approve output they can't see.
- **1.14.5 / 1.14.6:** it regressed **further, to a hard crash** — `human_input` now raises `AttributeError` before any prompt, because the new default experimental `AgentExecutor` is handed to the human-feedback provider but doesn't expose the `ask_for_human_input` the provider requires.

### Environment

Python 3.12; reproduced on crewai 1.14.4 and 1.14.6.

### Reproduction

Deterministic, no API key / network — run on each version:

```python
from crewai import Agent, Crew, Task, Process
from crewai.llms.base_llm import BaseLLM

class StubLLM(BaseLLM):
    def call(self, messages, tools=None, callbacks=None, available_functions=None,
             from_task=None, from_agent=None, response_model=None):
        return "Thought: I know it.\nFinal Answer: The sky is blue."
    def supports_function_calling(self) -> bool:
        return False

agent = Agent(role="Tester", goal="Answer", backstory="bg",
              llm=StubLLM(model="stub"), verbose=False)
task = Task(description="Sky colour?", expected_output="a colour",
            agent=agent, human_input=True)
Crew(agents=[agent], tasks=[task], process=Process.sequential, verbose=False).kickoff()
```

### Behaviour by version

- **1.14.4** — prints only the feedback panel; no result. Setting `verbose=True` makes the `✅ Agent Final Answer` panel appear, confirming the result display is verbose-gated. The operator is asked to review a "Final Result above" that was never shown.
- **1.14.6** — crashes before the prompt:

```
crewai/experimental/agent_executor.py, in invoke
    formatted_answer = self._handle_human_feedback(formatted_answer)
crewai/experimental/agent_executor.py, in _handle_human_feedback
    return provider.handle_feedback(formatted_answer, cast("ExecutorContext", self))
crewai/core/providers/human_input.py, in _handle_regular_feedback
    while context.ask_for_human_input:
AttributeError: 'AgentExecutor' object has no attribute 'ask_for_human_input'
```

### Root cause

- **Hidden result (1.14.4):** `crew_agent_executor.py::_show_logs` emits `AgentLogsExecutionEvent` with `verbose=agent.verbose or crew.verbose`, so the result render is verbose-gated — while the feedback gate fires on `human_input` alone. `core/providers/human_input.py::_prompt_input` prints a prompt that references "the Final Result above" but never the result itself.
- **Crash (1.14.5+):** the experimental `AgentExecutor` is `cast("ExecutorContext", self)` into the provider but doesn't implement `ask_for_human_input`, which `_handle_regular_feedback` reads (`while context.ask_for_human_input:`). The experimental executor doesn't satisfy the `ExecutorContext` protocol the human-input provider depends on.

### Expected

`human_input=True` should (a) work on the default executor, and (b) display the result under review regardless of `verbose`.

### Suggested fixes

1. Make the experimental `AgentExecutor` satisfy the `ExecutorContext` protocol (expose `ask_for_human_input`) — restores HITL on 1.14.5+.
2. When `human_input` is active, include the `AgentFinish` output in the feedback panel (or emit the result unconditionally) so the prompt never references invisible output.

---

coolhandle01:
> **Draft bug report #2 for the CrewAI repo** (`crewAIInc/crewAI`) — the *verbose-gating* half, split out from #6065 (the crash). Not about this PR; parked here to lift/paste upstream.

---

**Title:** `human_input=True`: the feedback prompt references a "Final Result above" that is never displayed unless `verbose=True`

### Description

When a `Task` has `human_input=True` but the agent/crew is **not** verbose, CrewAI shows the `💬 Human Feedback Required` panel — *"Provide feedback on the Final Result above"* — but the result is **never printed**. The operator is asked to approve output they cannot see.

The two behaviours are gated independently:
- the **result render** (`_show_logs` → `AgentLogsExecutionEvent`) is gated on `verbose=agent.verbose or crew.verbose`;
- the **feedback gate** fires whenever `human_input=True`.

So `human_input=True` + non-verbose ⇒ a prompt that explicitly references content ("the Final Result above") that was never rendered.

### Relationship to #6065

This is **separate** from #6065 (the `AttributeError: 'AgentExecutor' object has no attribute 'ask_for_human_input'` crash). On 1.14.5 / 1.14.6 that crash fires first and masks this. The bug here is observable on **1.14.4** (where `human_input` still works), and will resurface on 1.14.5+ once #6065 / #6069 lands — so it's worth fixing alongside it.

### Environment

Python 3.12; reproduced on `crewai==1.14.4`.

### Reproduction

Deterministic, no API key / network:

```python
from crewai import Agent, Crew, Task, Process
from crewai.llms.base_llm import BaseLLM

class StubLLM(BaseLLM):
    def call(self, messages, tools=None, callbacks=None, available_functions=None,
             from_task=None, from_agent=None, response_model=None):
        return "Thought: I know it.\nFinal Answer: The sky is blue."
    def supports_function_calling(self) -> bool:
        return False

agent = Agent(role="Tester", goal="Answer", backstory="bg",
              llm=StubLLM(model="stub"), verbose=False)
task = Task(description="Sky colour?", expected_output="a colour",
            agent=agent, human_input=True)
Crew(agents=[agent], tasks=[task], process=Process.sequential, verbose=False).kickoff()
```

Run it and press Enter at the prompt.

### Expected

When `human_input` is requested, the result under review is displayed (or embedded in the feedback panel) **regardless of `verbose`** — the prompt should never reference output that wasn't shown.

### Actual

- `verbose=False`: only the feedback panel appears; no result anywhere.
- `verbose=True`: the `✅ Agent Final Answer` panel renders, and the feedback panel includes a `Final Output:` line — confirming the result display is purely verbose-gated.

### Root cause

`crew_agent_executor.py::_show_logs` emits `AgentLogsExecutionEvent(verbose=agent.verbose or crew.verbose)`, so the result render is verbose-gated, while the human-feedback gate (`core/providers/human_input.py::_prompt_input`) fires on `human_input` alone and prints a prompt that assumes the result is "above".

### Suggested fix

When `human_input` is active, surface the `AgentFinish` output regardless of `verbose` — either emit the result event unconditionally in that path, or include the output directly in the feedback panel so "the Final Result above" is always actually present.

---

coolhandle01:
> **Tested patch for CrewAI #6072** (the verbose-gating bug) — for your `crewAIInc/crewAI` fork, not this repo. Developed against `crewai==1.14.4` and verified with the stub repro.

**Target file:** `src/crewai/core/providers/human_input.py` (drop the `src/` prefix if your checkout uses a flat layout). Apply with `git apply <file>.patch`.

**What it does:** threads the result under review into `_prompt_input` / `_prompt_input_async`, which render it (a `📋 Result for Review` panel) *inside the same paused-live block* as the feedback prompt — so "the Final Result above" is always actually there, independent of `verbose`. Covers re-prompts in the feedback loop (shows the latest answer each round) and the async path.

```diff
--- a/src/crewai/core/providers/human_input.py
+++ b/src/crewai/core/providers/human_input.py
@@ -179,7 +179,9 @@
         Returns:
             The final answer after feedback processing.
         """
-        feedback = self._prompt_input(context.crew)
+        feedback = self._prompt_input(
+            context.crew, self._get_output_string(formatted_answer)
+        )
 
         if context._is_training_mode():
             return self._handle_training_feedback(formatted_answer, feedback, context)
@@ -200,7 +202,9 @@
         Returns:
             The final answer after feedback processing.
         """
-        feedback = await self._prompt_input_async(context.crew)
+        feedback = await self._prompt_input_async(
+            context.crew, self._get_output_string(formatted_answer)
+        )
 
         if context._is_training_mode():
             return await self._handle_training_feedback_async(
@@ -261,7 +265,9 @@
             else:
                 context.messages.append(context._format_feedback_message(feedback))
                 answer = context._invoke_loop()
-                feedback = self._prompt_input(context.crew)
+                feedback = self._prompt_input(
+                    context.crew, self._get_output_string(answer)
+                )
 
         return answer
 
@@ -315,18 +321,22 @@
             else:
                 context.messages.append(context._format_feedback_message(feedback))
                 answer = await context._ainvoke_loop()
-                feedback = await self._prompt_input_async(context.crew)
+                feedback = await self._prompt_input_async(
+                    context.crew, self._get_output_string(answer)
+                )
 
         return answer
 
     # ── I/O ───────────────────────────────────────────────────────────
 
     @staticmethod
-    def _prompt_input(crew: Crew | None) -> str:
+    def _prompt_input(crew: Crew | None, result: str = "") -> str:
         """Show rich panel and prompt for input.
 
         Args:
             crew: The crew instance for context.
+            result: The result under review. Rendered above the prompt so the
+                "Final Result above" reference holds regardless of ``verbose``.
 
         Returns:
             User input string from terminal.
@@ -340,6 +350,15 @@
         formatter.pause_live_updates()
 
         try:
+            if result:
+                formatter.console.print(
+                    Panel(
+                        Text(result),
+                        title="📋 Result for Review",
+                        border_style="cyan",
+                        padding=(1, 2),
+                    )
+                )
             if crew and getattr(crew, "_train", False):
                 prompt_text = (
                     "TRAINING MODE: Provide feedback to improve the agent's performance.\n\n"
@@ -375,11 +394,13 @@
             formatter.resume_live_updates()
 
     @staticmethod
-    async def _prompt_input_async(crew: Crew | None) -> str:
+    async def _prompt_input_async(crew: Crew | None, result: str = "") -> str:
         """Show rich panel and prompt for input without blocking the event loop.
 
         Args:
             crew: The crew instance for context.
+            result: The result under review. Rendered above the prompt so the
+                "Final Result above" reference holds regardless of ``verbose``.
 
         Returns:
             User input string from terminal.
@@ -393,6 +414,15 @@
         formatter.pause_live_updates()
 
         try:
+            if result:
+                formatter.console.print(
+                    Panel(
+                        Text(result),
+                        title="📋 Result for Review",
+                        border_style="cyan",
+                        padding=(1, 2),
+                    )
+                )
             if crew and getattr(crew, "_train", False):
                 prompt_text = (
                     "TRAINING MODE: Provide feedback to improve the agent's performance.\n\n"
```

**Verification** (bare crew, stub LLM, `verbose=False`, on `1.14.4`):

```
╭──────── 📋 Result for Review ────────╮
│  The sky is blue.                    │
╰──────────────────────────────────────╯
╭────── 💬 Human Feedback Required ─────╮
│  Provide feedback on the Final Result above. ...
```

Before the patch that top panel was absent.

**Suggested test** (adapt to CrewAI's layout/fixtures — I couldn't run their suite here):

```python
def test_prompt_input_renders_result_above_prompt(monkeypatch):
    import io
    from rich.console import Console
    from crewai.core.providers.human_input import SyncHumanInputProvider
    from crewai.events.event_listener import event_listener

    buf = io.StringIO()
    monkeypatch.setattr(event_listener.formatter, "console", Console(file=buf, width=100))
    monkeypatch.setattr("builtins.input", lambda: "")

    SyncHumanInputProvider._prompt_input(crew=None, result="REVIEW ME 123")

    out = buf.getvalue()
    assert "REVIEW ME 123" in out          # the result is shown...
    assert "Result for Review" in out      # ...regardless of verbose
```

Wire `Fixes #6072` into the PR description.

---

h1-api.md
coolhandle01, line 1: This file should be removed, useful though it was.

coolhandle01:
## Next step (tomorrow): depend on `crewui`, drop the inline generic TUI

[`crewui` 0.1.0](https://pypi.org/project/crewui/) is now on PyPI — the generic `CrewAIPipelineTUI` extracted out of this branch's `tools/tui/`. So this branch can shed the generic layer and keep only the cybersquad-specific wrapper ("keep cybersquad cyber").

**What crewui ships:** `CrewAIPipelineTUI` plus the helpers (`_make_tui_human_input_provider`, `format_step_message`, `route_log_record`, `task_layout`, `truncate`, `_TUILogHandler`), a `default.tcss`, a CLI / `__main__`, and `py.typed`. Requires `crewai` + `textual`.

**Heads-up — the constructor evolved during extraction, so this is a rework, not a drop-in dep bump:**

```
# was (inline):  CrewAIPipelineTUI(crew, task_map, record_prefix, verbose, dry_run)
# now (crewui):  CrewAIPipelineTUI(crew, record_prefix='pipeline', pipeline_name='',
#                                  dry_run=False, on_start=None, on_complete=None,
#                                  get_token_cost=None)
```

`task_map`/`verbose` are gone; `pipeline_name`, the `on_start`/`on_complete` callbacks, and a `get_token_cost` hook are new.

**Plan:**

1. Add `crewui>=0.1.0` to `pyproject.toml`; drop the direct `textual` dep (crewui brings it) and the `tools.tui` package declaration.
2. Delete the inline generic code: `tools/tui/__init__.py`'s `CrewAIPipelineTUI`, `tools/tui/_helpers.py`, and `tools/tui/default.tcss`.
3. Rewrite the `CybersquadTUI` wrapper to subclass `crewui.CrewAIPipelineTUI` on the new signature — `record_prefix="cybersquad"`, wire cybersquad's token cost into `get_token_cost`, and move whatever `task_map`/`verbose` did into the callbacks / `pipeline_name`.
4. Re-run the "Before you commit" stack. `tests/test_tui_helpers.py` will need reworking — the helpers it imports now live in `crewui`, so those tests either move to crewui or become thin wrapper-wiring tests here.

**Open question for a real run:** whether the crewai fork pin is still needed once on crewui, or whether crewui's own log pane renders the result so the fork's Result-for-Review panel is moot inside the TUI. The fork doesn't hurt either way.

Leaving this here to handle tomorrow.

---

coolhandle01:
> Reviewer Claude

Review of #166 at `ecd443c`. Independent session. The `crewui` / crewai git pins are **excluded at the author's request** — they are known, deliberate, and tracked; the bump script is an intentional QA affordance. Everything below is the rest of the diff.

CI is green across all 9 checks. I verified the integration seam by construction rather than by reading: imported `crewui` at the pinned ref and built `CybersquadTUI` against it — the subclass composes, `CSS_PATH` resolves to two entries in the right order (crewui's `default.tcss` first, `cybersquad.tcss` second so it wins ties), and all seven kwargs at `main.py:161-169` match `CrewAIPipelineTUI.__init__`'s signature exactly. The adapter layer is correct.

---

## Major

### 1. No TTY guard on the default path, and `--headless` is undocumented

`main.py:_run_tui`.

Default invocation (`python main.py`, no `--headless`) unconditionally constructs and `.run()`s the Textual app. In a non-TTY context — CI, a piped or redirected shell, a daemon, a web session — Textual's `App.run()` raises, with no cybersquad-level catch and no message pointing the operator at the escape hatch.

The asymmetry is the tell: `_run_headless` has proper handling (`KeyboardInterrupt` → exit 0, `Exception` → exit 1). The TUI path has none.

Compounding it, the module docstring's "Usage" block documents `--verbose` and `--dry-run` but **not** `--headless`, so the one workaround is undiscoverable without reading `parse_args`.

There is also no code path at all for crewui being unimportable — `tools/tui/__init__.py` does a hard top-level `from crewui import CrewAIPipelineTUI`.

**Fix:** detect a non-interactive stdin/stdout and either fall back to headless automatically or exit with a clear "no TTY detected; re-run with `--headless`". Add `--headless` to the usage docstring.

---

## Minor

### 2. `bump-crewui-git-pin.sh` — two gaps worth closing while it lives

The script itself is sound: `set -euo pipefail`, exec bit set (`100755`), honest about being break-glass, and carries its own removal instruction. Two things undercut it in use:

- **`--no-deps` silently skips dependency changes.** If the crewui branch gains or moves a runtime dependency, the venv keeps the old set and you QA against a tree that does not match the branch's declared requirements. Not hypothetical: `psutil` landed on crewui `main` and reaches the PR branch via its merge commits, and #4's `TextArea(placeholder=...)` needs `textual>=6.0` while crewui still declares `>=0.60`. A `--no-deps` refresh will not surface either.
- **The verification line cannot verify anything.** `pip show crewui | grep -E '^(Name|Version)'` prints `0.2.0` for crewui `main` *and* for every commit on `feat/review-gate-ux` — `__version__` has not moved. So the "now installed:" output is identical whether the refresh worked or silently no-oped. The resolved commit is recorded by pip in the dist-info `direct_url.json` (PEP 610, `vcs_info.commit_id`); reading that gives a check that can actually fail.

Suggested: drop `--no-deps` (or offer it as an opt-in fast path), and print the resolved SHA:

```bash
python -c "import json,importlib.metadata as m,pathlib; \
p=pathlib.Path(m.distribution('crewui')._path)/'direct_url.json'; \
print(json.load(p.open())['vcs_info']['commit_id'])"
```

### 3. `test_current_opus_pricing_does_not_collide_with_legacy` does not pin what it claims

`tests/tools/test_metrics.py:38-41` documents the `max(..., key=len)` longest-prefix tie-break in `estimate_cost` (`tools/metrics.py:53-70`). It does not defend it: because `_PRICING` happens to list the specific keys (`claude-opus-4-5`) *before* the legacy `claude-opus-4`, a naive first-match loop returns the same $30 for every input in the test. A regression back to first-match iteration keeps the whole file green.

**Fix:** add a case where a shorter key that is a prefix is inserted *first*, or assert the resolution is order-independent.

### 4. `TestDryRunSummary.test_renders_agents_and_tasks` asserts a crash, not output

`tests/test_main.py:56-77`. It exercises both the tools/no-tools and human-review branches but asserts only `"DRY RUN" in ...out`. Any row-rendering regression passes. Assert that a role and a tool name actually appear.

### 5. The coverage `omit` of `tools/tui/__init__.py` rests on an inaccurate premise

`pyproject.toml:165-170` justifies the omit as "a bare subclass with no logic of its own." Its `CSS_PATH` list construction is load-bearing — crewui base first, `cybersquad.tcss` last, with a deliberate re-widening of the type — and `tests/test_tui.py` explicitly asserts that ordering. The module is imported and exercised by the tests anyway, so the omit needlessly drops a tested file from the denominator.

### 6. `_run_headless` accepts `verbose` and never uses it

`main.py:172`. The parameter is accepted, asserted by `test_headless_dispatches_to_run_headless`, and never read in the body — so the test pins wiring to an ignored argument. (Crew verbosity does still apply via `build_crew(verbose=...)`, so this is dead wiring rather than a lost feature. Likely ruff `ARG001` territory.)

Related and benign: `--verbose` has no effect on TUI chrome because `CrewAIPipelineTUI.__init__` has no `verbose` parameter at all — it was dropped during the crewui extraction. Worth a docstring note that `--verbose` is a headless-only knob.

### 7. New "exits nonzero, drops data" path in the metrics move

`tools/metrics.py:121` changes `save_metrics(metrics, reports_dir: str)` to `save_metrics(metrics, run_dir: Path)`, moving output from `reports/<run_id>/metrics.json` to `<run_dir>/metrics.json`. Callers are updated (`main.py:156, 215`) and the new path is asserted (`test_metrics.py:93-95`) — the change is fine.

The edge: a headless run that returns `token_usage` but where the PM never bound a programme now hits `runtime.run_dir()`'s `RuntimeError`, which the surrounding `except Exception` swallows into `sys.exit(1)`. Metrics are silently lost where the old `reports_dir/<run_id>` path would have saved them. Degenerate (a run with no selected programme), but it is a new failure mode with no test.

### 8. Dead-code removal is unrelated scope

`runtime.py` drops `programme_cache_path()` and `tests/test_runtime.py` its two tests. To be clear, since the shape of this deletion looks alarming: **no invariant was removed.** `bind_run_id`, `bind_programme`, `run_dir`, and the `#128` conflict-guard are all intact at HEAD, and `programme_cache_path` was a stateless path builder with zero production callers on `main` before this branch — verified. Legitimate cleanup, just cleanup that belongs in its own `chore/` branch under "one PR does one thing".

---

## Carried over from crewui #4

Reviewed separately (coolhandle01/crewui#4), but two land squarely on this consumer:

- **Unescaped agent output raises `MarkupError` in crewui's session pane**, and the accumulating-buffer rewrite makes one bad message poison every subsequent write in that turn. Verified: `Text.from_markup("see file [/etc/hosts]")` raises. This repo is the reason it matters — a pentest pipeline emits bracketed paths as ordinary output, so this will fire in normal operation, not at an edge.
- **A duplicate `Submitted` can auto-answer the next review gate with the previous round's text.** A human-approval gate on a tool that acts against live targets is the wrong place for a lost keystroke.

Neither is a cybersquad defect; both gate the value this PR delivers.

---

## Sound

- **`crew.py` `Memory | None` → `Memory | bool`** — correct and correctly cited: `False`, not `None`, is CrewAI's disabled sentinel, and the docstring names the telemetry span-attribute rejection of `NoneType`. Disabled path is tested.
- **`build_crew` as a `@contextmanager` over a pure `_assemble_crew`** — sound. Provisioning is unconditional so dry-run still provisions, matching the `cybersquad-mcp` build-time-only rule, and the `with` block in `main()` correctly encloses both surfaces.
- **`tests/conftest.py`** — the additions are module-level import-time code, not an autouse fixture: `REPORTS_DIR` setdefault plus a loop clearing 17 scan-tuning env vars before `config.py`'s singleton loads. Running once at import and only *clearing* operator dials to a CI baseline, it cannot mask a per-test failure or leak state between tests. `./test-results` is gitignored. The hermeticity rationale is well argued.
- **The new tests are real, not mock theatre.** `TestRunHeadless` asserts observable behaviour across dry-run, no-usage, usage, `KeyboardInterrupt`→0 and `Exception`→1. `TestRunTui` invokes the captured `on_start` / `on_complete` / `get_token_cost` closures and asserts persistence plus the run-id-must-not-leak-to-TUI contract. `TestMain` uses a real `@contextmanager` fake matching production's `with build_crew(...)` scope. `test_theme_layer_thins_and_aligns_scrollbars` is a genuine headless Textual render measuring real scrollbar geometry — it would catch the padding-on-container regression it documents.
- **Shared fixtures respected** — no redefinition of `tests/fixtures/` equivalents. The `monkeypatch.setattr(runtime, "run_dir", ...)` deviation is commented and defensible (`save_metrics` is mocked, so the value is never used for I/O).
- **Secrets and config hygiene clean** — `.env.example` adds only telemetry opt-out vars with no values; `.gitignore` additions only widen ignores; the `default_factory=lambda` safety invariant is preserved. `.claude` skill edits are a consistent model-name refresh, ASCII-clean.

The wrapper is small, the seam is correct, and the tests around it are honest. The Major above is the one thing I would want closed before this is exercised anywhere without a terminal.

---

PR #171, opened 2026-05-26
feat(mcp): provision time MCP at build_crew() + cybersquad-mcp skill (#144)

coolhandle01:
## Summary

Provisions Model Context Protocol servers at `build_crew()` time, behind a small package, with the discipline the MCP spec asks for at the trust boundary. Ships with the time MCP as the first server.

## The discipline

The MCP spec (2025-06-18, *Security and Trust & Safety*) treats tool names, descriptions, and inputSchemas as untrusted unless from a trusted server — and CrewAI surfaces all of them into the agent's prompt the moment a server is listed, not just when a tool is called. Five rules, codified in [`.claude/skills/cybersquad-mcp/`](./.claude/skills/cybersquad-mcp/SKILL.md) so future contributors land on them:

1. Vendor package pinned exactly; consumed through its typed models.
2. `MCPServerAdapter` materialised only inside `mcp_servers/`.
3. Provisioned and discovered MCP sets stay disjoint.
4. Explicit `*tool_names` allowlist per MCP — a version bump that adds a tool can't silently widen the agent's surface.
5. Two-line audit log per adapter (intent before, resolved after).

## Shape

```
mcp_servers/
├── __init__.py     provisioned_mcp_tools() orchestrator + ProvisionedMCPTools registry
├── _common.py      shared utilities (adapter-stack pre-flight)
└── _time.py        the time MCP (one submodule per provisioned MCP)
```

`crew.py` distributes the registry crew-wide via `build_agent(crew_wide_mcp_tools=...)`. `main.py` wraps `crew.kickoff()` in the context manager; dry-run skips MCP startup. Each MCP has a `<name>_enabled` boolean in `MCPConfig` defaulting `false`; flipping it on requires `pip install -e ".[mcp]"`.

## Test plan

- [ ] `ruff check . && ruff format --check .` green
- [ ] `mypy . --ignore-missing-imports` green
- [ ] `pylint .` ≥ 9.85
- [ ] `bandit -c pyproject.toml -r .` no findings
- [ ] `pytest` green, 100% diff coverage on touched `.py` files
- [ ] Live MCP startup: `CYBERSQUAD_MCP_TIME_ENABLED=true python -c "from mcp_servers import provisioned_mcp_tools; ..."` lists both allowed tools and exits clean


squad/__init__.py
```diff
@@ -155,21 +155,34 @@ def skills_dir(self) -> Path:
         return self.dir / "skills"
 
 
-def build_agent(member: SquadMember, llm: object, verbose: bool = False) -> Agent:
+def build_agent(
+    member: SquadMember,
+    llm: object,
+    verbose: bool = False,
```
coolhandle01, line 161: Does this have to be `object`? Also can `Any` be tightened?

squad/__init__.py
```diff
@@ -155,21 +155,34 @@ def skills_dir(self) -> Path:
...
+    tests; MCP tools come from a third-party adapter and live outside
+    the ``SquadTool`` Protocol surface by design.
     """
     skills: list[Path] = [member.skills_dir] if member.skills_dir.is_dir() else []
     return Agent(
         role=member.read("role"),
         goal=member.read("goal"),
         backstory=member.read("backstory"),
-        tools=member.tools,
+        tools=[*member.tools, *extra_tools],
```
coolhandle01: I feel like there should be a bit more ceremony around injecting the MCP tools (I think that's what's happening here?)

mcp_servers.py
coolhandle01, line 1: This is already worth promoting to a package mcp/ I think, because as soon as we add another MCP or two it'll be too long..

mcp_servers/_time.py
```diff
@@ -0,0 +1,94 @@
...
+    Requires both the vendor server module (``mcp_server_time``) and the
+    CrewAI MCP adapter stack (``mcpadapt`` via ``crewai-tools[mcp]``).
+    If anything is missing the package-level warning fires and the
+    server is skipped, mirroring the missing-binary pattern.
+    """
+    return mcp_adapter_stack_usable() and importlib.util.find_spec("mcp_server_time") is not None
+
+
+def enter(stack: ExitStack) -> Sequence[BaseTool]:
+    """Materialise and start the time MCP, returning its allowed tool list.
```
coolhandle01, line 81: Are we not using https://github.com/modelcontextprotocol/servers/tree/main/src/time ?

coolhandle01: Yes — `mcp-server-time==2026.1.26` in the `[mcp]` extra is that server's PyPI publication. f1bdc08 surfaces the source link in the module docstring and next to the pin so the connection's visible from both.

---

Issue #176, opened 2026-05-27
feat: flesh out nmap recon tool - modes, evidence, scan-mode-aware, async


coolhandle01:
## Motivation

`tools/recon/nmap.py` today is `nmap -sS --open -T4 -F -oG -` - top-100 TCP SYN scan, ports only, no service/version, no banners, no scan-mode awareness. It is the only major recon tool in the squad that does NOT consult `config.scan.scan_mode` (STEALTH / NORMAL / RAID), and the OA cannot ask it to do anything richer than "list open ports."

The httpx tech-detect pass at `tools/recon/probe.py:29-44` is currently the primary source of framework / version signal (populating `Endpoint.technologies`). Nmap with `-sV` and `--script=banner,http-server-header,http-title` would be the second source, particularly for non-HTTP services (SSH banners, MySQL handshakes, Redis info) that httpx cannot see.

Split out of #83 (framework-aware PT) - the framework refactor consumes existing recon signal; this issue grows the recon signal itself.

### 1. `NmapMode` StrEnum + scan-mode-aware flag assembly

```python
class NmapMode(StrEnum):
    QUICK_PORTS = "quick-ports"          # today's behaviour: -sS -F
    SERVICE_VERSION = "service-version"  # adds -sV
    FULL_INVENTORY = "full-inventory"    # -sV + --script=banner,http-server-header,http-title
    OS_DETECT = "os-detect"              # -O (privileged-only; refuse cleanly if unprivileged)
```

The OA picks a `NmapMode` in agent-prose terms; the wrapper composes the final flag list from `(NmapMode, config.scan.scan_mode)`. Examples:

| NmapMode | STEALTH | NORMAL | RAID |
|---|---|---|---|
| `QUICK_PORTS` | `-sS -F -T2` | `-sS -F -T3` | `-sS -F -T4 --min-rate 1000` |
| `SERVICE_VERSION` | `-sV --version-intensity 2 -T2` | `-sV --version-intensity 5 -T3` | `-sV --version-intensity 9 -T4 --min-rate 1000` |
| `FULL_INVENTORY` | adds `--script=banner,http-server-header,http-title -T2` | same scripts, `-T3` | same scripts, `-T4 --min-rate 1000` |
| `OS_DETECT` | `-O -T2` | `-O -T3` | `-O -T4` |

Two knobs (mode + scan_mode), no flag-construction in the LLM.

### 2. Widen to `hosts: TargetHostnames`

[...]

Issue #88, opened 2026-05-15
Exploit interface: typed, composable, precondition-gated attack primitives


coolhandle01:
## Why this, not a kill chain

Lockheed Martin&#39;s [cyber kill chain](https://www.lockheedmartin.com/en-us/capabilities/cyber/cyber-kill-chain.html) is linear by design: Recon → Weaponize → Deliver → Exploit → Install → C2 → Actions on Objectives. The linearity leaks into how you build an agent on top of it. A &#34;next step&#34; framing implies the agent picks one, which implies the agent hallucinates one when the data does not say.

Real adversaries do not walk a chain. They branch, abandon, pivot, and search for the **path of least resistance** — the same shortest-cost-path problem solved on attack graphs with Dijkstra-class algorithms, and the same shape games-AI work on bots in Quake / ioDoom solves with A\* over weighted graphs, predictive collision avoidance under cost, Bayesian opponent modelling.

Academic backing, in two arcs that converge:

**Attack-graph formalism over kill-chain narrative:**

- Schneier, B. (1999). [&#34;Attack Trees.&#34;](https://www.schneier.com/academic/archives/1999/12/attack_trees.html) *Dr. Dobb&#39;s Journal*. — canonical lay-accessible treatment.
- Phillips, C., Swiler, L. P. (1998). [&#34;A Graph-Based System for Network-Vulnerability Analysis.&#34;](https://dl.acm.org/doi/10.1145/310889.310919) *NSPW &#39;98*. — first to frame attacker progression explicitly as shortest-path / minimum-cost search over an attack graph with arc weights for effort and probability of success.
- Sheyner, O., Haines, J., Jha, S., Lippmann, R., Wing, J. M. (2002). [&#34;Automated Generation and Analysis of Attack Graphs.&#34;](https://ieeexplore.ieee.org/document/1004377) *IEEE Symposium on Security and Privacy*. — automated graph generation; the symbolic-model precedent for what amass + the label system here build dynamically.

**Path-finding-as-graph-search lineage:**

- Hart, P. E., Nilsson, N. J., Raphael, B. (1968). [&#34;A Formal Basis for the Heuristic Determination of Minimum Cost Paths.&#34;](https://ieeexplore.ieee.org/document/4082128/) *IEEE Transactions on Systems Science and Cybernetics, 4(2), 100–107*. — origin of A\*; the heuristic that lets a graph-search agent prefer high-reward / low-cost subtrees over uniform breadth-first exploration.
[...]

Issue #75, opened 2026-05-15
XXE via SVG upload endpoints


coolhandle01:
## Summary

SVG is XML. Server-side SVG processing (thumbnail generation, PDF export, ImageMagick/Inkscape conversion) expands XML entities before rendering, making it a distinct XXE attack surface separate from the generic `application/xml` probe (#49).

## Relation to existing work

Extends #74 (SAML XXE) and #49 (generic XXE probe). SVG is the third XML dialect worth probing systematically - generic XML/SOAP/XML-RPC all POST to API endpoints, while SVG XXE targets file upload flows.

## Attack surface

Endpoints accepting `image/svg+xml` or multipart uploads where the filename extension is `.svg` and a server-side image processor is likely (avatar upload, logo upload, document export, PDF generation, preview generation).

### Tool location

`tools/pentest/svgxxe.py` or as an extension of `tools/pentest/xxe.py`.

### SVG XXE payload

```xml
<?xml version="1.0"?>
<!DOCTYPE foo [<!ENTITY xxe SYSTEM "file:///etc/passwd">]>
<svg xmlns="http://www.w3.org/2000/svg">
  <text>&xxe;</text>
</svg>
```

### Detection strategy

1. Identify candidate upload endpoints from recon: URLs containing `upload`, `avatar`, `image`, `logo`, `import`, `convert`, endpoints with multipart form fields.
2. POST multipart payload with `filename="probe.svg"` and SVG XXE body.
3. Match `/etc/passwd` or `win.ini` sentinel markers in response body -> CRITICAL.
4. Optionally: check for XML parser error markers if no file read lands -> MEDIUM.

### Challenges

- Requires multipart form construction (`requests` handles this natively).
- No guarantee the server reflects SVG content in the response - may need to check a separate "preview" or "thumbnail" URL, which complicates in-band detection.
- If output is not reflected, falls back to OOB (blocked pending callback server infrastructure).

## Acceptance criteria

- [ ] Craft a valid SVG document with embedded XXE SYSTEM entity
- [ ] POST as multipart to candidate upload endpoints identified by recon
- [ ] Match sentinel markers in response -> CRITICAL
- [ ] Fall back to XML error marker detection -> MEDIUM
- [ ] Unit tests mock all HTTP; no real file upload required
- [ ] `bandit`, `ruff`, `mypy` all pass

PR #177, opened 2026-05-28
feat(#83): typed Framework / Cloud / Service stamps + Webpage class

### LINKED ISSUE #83: feat: tools/pentest/framework/ - framework-specific vulnerability probes

coolhandle01:
Closes #83.

## Summary

- **Three asset-property StrEnums** typed in `models/` — `Framework` (django, rails, spring, angular, tornado, laravel, nextjs), `Cloud` (aws, azure, gcp), `Service` (redis, mongodb, elasticsearch, couchdb, mysql, postgresql, grafana, kibana, portainer, cpanel, plesk, directadmin, webmin, consul, vault) — surfaced into the agent's tool descriptions via four sibling classification decorators (`@owasp` / `@framework` / `@cloud` / `@service`) that stack onto each `check_X` function.
- **`Webpage` class in `tools/html.py`** encapsulates the soup recipes inlined in `csrf.py` (forms parsing with resolved actions, CSRF cookie/meta detection, framework signals). Probes consume a typed page object instead of `BeautifulSoup` mechanics. `html.fetch()` now returns `Webpage`; CSRF probe is pure orchestration; `sri.py` + `sourcemaps.py` mechanically migrated.
- **`AttackPlanItem.owasp_category: OWASPCategory | None`** lets the VR cite the OWASP code per plan item; the PT pattern-matches against probe tools' `@owasp` stamps for selection. `OWASPCategory` promoted to `models/owasp.py` (re-exported from `tools/pentest/owasp.py` so 27 existing consumers are unchanged) to dodge a package-level circular import.
- **`@cloud(...)` stamps** applied to S3 + Azure probes. **`@service(...)` stamps** applied to 14 product-specific probes (databases / dashboards / hosting control panels / service-discovery).
- **Vocabulary system** in `tools/research_vocab.py` composes `Framework` / `Cloud` / `Service` members into the Vulnerability Researcher's `Finalise Research` docstring; PT `description.md` decision-guidance prose cites typed members in operational voice (no Python jargon, no function-name leaks).
- **Wappalyzer docs link** in `tools/recon/probe.py` so the next contributor sees where the framework strings come from.

## Design notes

The four classification decorators (`@owasp` / `@framework` / `@cloud` / `@service`) are a **stable layer**, not interim plumbing. `#88`'s `@attack(name=, requires={Label.X, ...})` proposal builds *above* this layer (label-gated exploit selection), not in place of it. The interim-#88 FIXMEs that landed in earlier PRs have been stripped accordingly.

The probe signature reshape (`check_X(endpoint: Endpoint, page: Webpage) -> RawFinding | None`) discussed during the design phase is deferred — it's a 10-probe cross-cutting refactor of the whole `tools/pentest/` surface and warrants its own issue.

## Test plan

- [x] 1689 unit tests pass (baseline on `main` was 1651; +38 covering the new decorators, `Webpage` shape, and `AttackPlanItem.owasp_category`)
- [x] `ruff check .` passes
- [x] `ruff format --check .` passes
- [x] `mypy .` clean (199 -> 204 source files)
- [x] `bandit` clean (3 pre-existing LOW findings, no new ones)

## Related

- #88 — label-gated exploits / typed `Exploit` interface (the layer above these stamps)
- #176 — nmap modes + `Framework` emission from banners (consumes the `Framework` enum landed here)
- `models/asset.py:LlmEndpoint` FIXME — `DiscoveredMCP` promotion remains foreseen but out of scope here

---
_Generated by [Claude Code](https://claude.ai/code/session_01KKtsvHiZsuWst3HYsJquc6)_


tests/tools/pentest/test_cloud_decorator.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test_azure_blob_containers_stamped_azure(self) -> None:
+        from tools.cloud.azure import check_azure_blob_containers
+
+        assert check_azure_blob_containers.clouds == (Cloud.azure,)  # type: ignore[attr-defined]
+
+    def test_azure_sas_tokens_stamped_azure(self) -> None:
+        from tools.cloud.azure import check_azure_sas_tokens
+
+        assert check_azure_sas_tokens.clouds == (Cloud.azure,)  # type: ignore[attr-defined]
```
coolhandle01: There's a lot of `# type: ignore[attr-defined]` here! We must be holding it wrong or doing something bad

tests/tools/test_html.py
```diff
@@ -6,72 +6,240 @@
...
         assert h1.text == "Hello"
 
-    def test_empty_soup_for_non_html_content_type(self, target_apex) -> None:
-        with patch(
-            "requests.get",
-            return_value=_mock_get(body='{"key": "val"}', content_type="application/json"),
-        ):
-            _, soup = html.fetch(f"https://{target_apex}/api")
+    def test_empty_soup_for_non_html_content_type(self, target_apex, make_response) -> None:
+        resp = make_response(body='{"key": "val"}', headers={"Content-Type": "application/json"})
```
coolhandle01:
```suggestion
        resp = make_response(body='{"key": "val"}', headers=_JSON_CT)
```

tests/tools/test_html.py
```diff
@@ -6,72 +6,240 @@
...
+
+        assert page.response is mock_resp
+
+
+class TestWebpageCsrfSignals:
+    """Cookie- and meta-tag-based CSRF protection detection."""
+
+    def _page(self, soup_html: str = "", cookies: dict | None = None) -> html.Webpage:
+        from bs4 import BeautifulSoup
+
```
coolhandle01: Can't use make_response()?

tests/tools/test_html.py
```diff
@@ -6,72 +6,240 @@
...
+class TestWebpageForms:
+    """Form parsing on the Webpage shape."""
+
+    def _page(self, soup_html: str, page_url: str = "https://example.com/") -> html.Webpage:
+        from bs4 import BeautifulSoup
+
+        resp = MagicMock()
+        resp.cookies = {}
+        resp.url = page_url
+        return html.Webpage(response=resp, soup=BeautifulSoup(soup_html, "html.parser"))
```
coolhandle01: Should resp.body also be soup_html..?

Three definitions of this function, which is lucky as I didn't spot this last time I saw one ;)

tests/tools/test_html.py
```diff
@@ -6,72 +6,240 @@
...
+            "</form>"
+        )
+        page = self._page(soup_html)
+        assert not page.post_forms[0].has_csrf_input
+
+    def test_forms_cached(self) -> None:
+        page = self._page('<form method="post" action="/x"></form>')
+        first = page.forms
+        second = page.forms
+        assert first is second
```
coolhandle01: This file is very long and the three classes are very similar. I feel like pytest parameterized stuff would help with different the soup_html inputs, more use of existing or introduction of new shared fixtures would help where necessary to keep things easy to follow

tools/cloud/databases/sql.py
coolhandle01, line 86: I'm not sure if MariaDBs are more or less vulnerable that MySQLs, generally speaking. Would VR/PT want to know which?

tools/pentest/cloud.py
```diff
@@ -0,0 +1,46 @@
+"""Cloud-provider targeting stamp for pentest check functions.
+
+Sibling to ``tools/pentest/owasp.py`` and ``tools/pentest/framework.py``.
```
coolhandle01, line 3: Should these stamps be defined in like, some more central place? Things are starting to "depend on the pentester"

tools/pentest/csrf.py
```diff
@@ -40,109 +41,28 @@
 _EVIL_REFERER = "https://evil.example.com/attack"
```
coolhandle01, line 41: This should be the cybersquad's oob server from config.py

tools/pentest/csrf.py
```diff
@@ -40,109 +41,28 @@
...
     """
     Detect CSRF vulnerabilities across HTML endpoints.
 
     Tier 1 (MEDIUM): GET each endpoint.  Skip non-HTML responses and any
-    page that signals framework-level CSRF protection via a CSRF-pattern
-    cookie (Angular, Django, Tornado) or a <meta name="csrf-token"> tag
-    (Rails, Spring).  For remaining pages, flag POST forms that contain no
-    hidden input whose name matches a CSRF-token pattern.
+    page where ``page_protected`` is True (CSRF-pattern cookie or meta
+    tag present - Angular, Django, Tornado, Rails, Spring).  For
```
coolhandle01, line 58: But no framework stamps? (Says me, python noob)

tools/html.py
```diff
@@ -1,29 +1,219 @@
...
+                    method=method,
+                    has_csrf_input=has_csrf,
+                )
+            )
+        return out
+
+    @cached_property
+    def post_forms(self) -> list[Form]:
+        """All <form method="post"> elements on the page."""
+        return [f for f in self.forms if f.method == "post"]
```
coolhandle01, line 307: Do people ever do put or patch forms?

tools/pentest/sourcemaps.py
```diff
@@ -129,10 +129,10 @@ def check_js_source_maps(endpoints: list[Endpoint]) -> list[RawFinding]:
             logger.debug("Source map: failed to fetch %s: %s", ep.url, exc)
             continue
 
-        if soup.find() is None:
+        if page.soup.find() is None:
             continue
 
-        for tag in soup.find_all("script", src=True):
+        for tag in page.soup.find_all("script", src=True):
```
coolhandle01: Maybe Webpage can do more of this work? I'm thinking like a get_scripts() and get_stylesheets().. having said that, "guessing" the .map URL should be outside the remit of Webpage, but Webpage should encapsulate soup usage mostly..

We might even need a class SubResource with a bool integrity or something.. 🤔

tools/html.py
```diff
@@ -1,29 +1,219 @@
...
+    that need raw access; derived page properties are
+    ``cached_property`` so a probe asking for ``post_forms`` and
+    ``frameworks`` only pays the parse cost once.
+
+    Not a frozen dataclass: ``cached_property`` writes to the instance.
+    Consumers should treat the instance as immutable - tests construct
+    a fresh Webpage per case rather than mutating a shared one.
+    """
+
+    def __init__(self, response: requests.Response, soup: BeautifulSoup) -> None:
```
coolhandle01: Isn't soup initialised from `response`? Do we need a soup in the ctor? The class should encapsulate soup I thought?

tools/html.py
```diff
@@ -1,29 +1,219 @@
...
+        self.response = response
+        self.soup = soup
```
coolhandle01:
```suggestion
        self.soup = BeautifulSoup(response.body)
```

Maybe I missed something but I was surprised to see soup being passed in

tools/pentest/cloud.py
coolhandle01: Sharp call - 27 imports across the pentest probes target `tools/pentest/owasp` / `framework` / `cloud` / `service`, so cloud / database / dashboard probes "depending on the pentester" is exactly the shape that's accumulating.

The enums themselves already live in `models/` (`models/framework.py` / `cloud.py` / `service.py`, plus `OWASPCategory` which got promoted to `models/owasp.py` in 9baecc0 to dodge a circular import). The decorator bodies are tiny wrappers over `getattr` / `__doc__` mutation - moving them next to their enums (or into a sibling `models/classifications.py` / `models/classifications/` package) is mechanical and lets `tools/cloud/aws.py` etc. import the stamp from `models` instead of from `tools/pentest`.

Two shapes that work:

**A) One file per stamp pair:** `@framework(...)` lands in `models/framework.py` alongside the `Framework` enum (and same for cloud / service / owasp). Each enum file already has the obvious docstring "the typed shape downstream agents read"; folding the decorator in keeps the read+write surface together.

**B) Sibling module:** `models/classifications.py` carrying all four stamps + `stamps_of()`. Reads import enums from their existing files; everything decorator-y lives in one place.

`tools/pentest/owasp.py` can either go away (cleanest) or stay as a re-export shim (zero-touch for the 27 existing imports). I'd lean **A + shim** so the existing imports keep working and we don't churn 27 files in this same PR. If you'd rather have the churn now, (B + no shim) is the alternative.

Want me to land that in this PR, or split it into a follow-up?

---

tools/pentest/csrf.py
coolhandle01: Investigated and there isn't one today - I greped `config.py` for `oob` / `callback` / `attacker` / `evil` / `controlled` and found nothing. Both this probe and `tools/pentest/cors.py` carry the same hardcoded `_EVIL_ORIGIN = "https://evil.example.com"` literal.

The cleanest move is a new `AppConfig.attacker_origin: str` (default `https://evil.example.com` so behaviour is unchanged for tests / operators who haven't set anything), then migrate CSRF + CORS to read `config.attacker_origin`. That gives operators a single env knob to point at a domain they actually control (matching the traceable-UA story in #46), and is the natural seam to grow into a full OOB callback server config when blind-SSRF / blind-XXE land.

Want me to land that in this PR or pull it into a follow-up issue? It's a tiny diff (3 files) but it does cross out of #83's scope a bit.

---

tools/pentest/csrf.py
coolhandle01: Deliberate - this came up during the design discussion. The stamps mean "this probe **specifically targets** framework X", not "this probe is aware of X exists." CSRF is framework-*aware* (it consumes `page.frameworks` / `page.page_protected` at runtime to suppress noise on JS-protected pages) but framework-*generic* (it fires across any HTML endpoint regardless of stack). Stamping it `@framework(django, rails, spring, ...)` would muddy that distinction; the real per-framework probes (#83's future `tools/pentest/framework/django.py` etc.) are when stamps start earning their keep.

The intent is documented in `tools/pentest/framework.py:framework`'s docstring ("Use ONLY on probes that specifically target one or more named frameworks ... do NOT stamp on framework-aware-but-generic probes (CSRF, headers)"), so future me reading the next probe knows the rule.

---

tools/html.py
coolhandle01: HTML spec only allows `get` and `post` natively for `<form method>` - any value the parser doesn't recognise gets coerced to `get`. PUT / PATCH / DELETE forms exist only as a server-side convention: Rails ships a `<input type="hidden" name="_method" value="patch">` field, Laravel has the `@method('PUT')` Blade directive, etc. At the wire level those still POST; the framework reads the hidden override.

So `post_forms` / `get_forms` is the complete coverage. Probes that care about the override semantics can look at the hidden inputs on a `Form` themselves (the `<input type="hidden" name="_method">` would show up in the same parse pass we already do for `has_csrf_input`).

Documented in the new `Form` docstring in 9baecc0 so future me sees the rule.

---

tools/html.py
```diff
@@ -84,9 +113,20 @@ class Webpage:
...
+    def __init__(self, response: requests.Response) -> None:
         self.response = response
-        self.soup = soup
+
+    @cached_property
+    def soup(self) -> BeautifulSoup:
+        """Parsed soup for the response body.
+
+        Returns an empty soup when the Content-Type is not text/html -
+        callers can treat an empty soup as a skip condition rather than
```
coolhandle01, line 177: Ahh fine let's do a page.mimetype(), then the caller can `match`/`case` on the return value.

tools/html.py
```diff
@@ -178,6 +218,51 @@ def get_forms(self) -> list[Form]:
...
+        for tag in self.soup.find_all("script", src=True):
+            src = str(tag["src"])
+            resolved = urljoin(self.url, src) if src else self.url
+            has_integrity = bool(tag.get("integrity"))
+            out.append(SubResource(url=src, resolved_url=resolved, has_integrity=has_integrity))
+        return out
+
+    @cached_property
+    def javascripts(self) -> list[SubResource]:
+        """Subset of ``scripts`` whose URL path ends in ``.js``.
```
coolhandle01: Does this need to care about .jsx or anything whacky like that to catch all the js?

tests/tools/pentest/test_sourcemaps.py
```diff
@@ -308,6 +313,7 @@ def test_skips_map_when_gitleaks_missing_and_no_internal_paths(
 
         def fake_get(url, **kwargs):
             resp = MagicMock()
```
coolhandle01: These should all probably be using the make_response() fixture

tests/tools/test_html.py
```diff
@@ -11,8 +11,31 @@
 
 pytestmark = pytest.mark.unit
 
-
 _HTML_CT = {"Content-Type": "text/html; charset=utf-8"}
```
coolhandle01, line 18: It occurs to me that `requests` or python itself must have definitions for these already, surely? 🤔🙏

tools/html.py
```diff
@@ -201,19 +286,17 @@ def _detect_frameworks(response: requests.Response, soup: BeautifulSoup) -> set[
...
+    (``frameworks``, ``page_protected``, ``forms`` / ``post_forms`` /
+    ``get_forms``, ``scripts``, ``stylesheets``). When the Content-Type
+    is not text/html the soup is empty and the derived properties return
+    their natural empty values.
     """
-    resp = http.get(url, **kwargs)  # type: ignore[arg-type]
-    ct = resp.headers.get("Content-Type", "")
-    body = resp.text if "text/html" in ct else ""
-    return Webpage(response=resp, soup=BeautifulSoup(body, "html.parser"))
+    return Webpage(http.get(url, **kwargs))  # type: ignore[arg-type]
```
coolhandle01, line 411: Please extend the linter ignore comment with a short justification

tools/html.py
coolhandle01: Checked - the stdlib `mimetypes` module guesses media types from file extensions (`mimetypes.guess_type("foo.html")` → `"text/html"`) but there's no constants table. `requests` ships none. `werkzeug` / `httpx` have parsers / classes but also no constants. So no canonical source.

The honest options:
1. **Keep the local `_HTML_CT` / `_JSON_CT` constants** (current state) — clear, local, scoped to one test file.
2. **Inline the dicts** at each call site — slightly noisier but no indirection.
3. **Promote to a fixture** in `tests/fixtures/responses.py` if the same pair shows up in another test file (no second consumer today).

I'd stick with (1) for now — they're four-line module-level constants used only in `test_html.py` and `test_sourcemaps.py` (where the `tools/pentest/sourcemaps.py` consumer also needs them, addressed in 0fd9261). If a third test file wants them, promote then.

---

tests/tools/test_html.py
coolhandle01: https://docs.python.org/3/library/mimetypes.html
.. are incorrect mime types a finding lol? Can incorrect mime type be abused? 

Anyways, this mimetypes library, it doesn't have the canonical list of mime types for us?

tests/tools/test_html.py
```diff
@@ -6,72 +6,371 @@
...
+    def test_response_is_returned_unchanged(self, target_apex, make_response) -> None:
+        mock_resp = make_response(body="<p>content</p>", headers=_HTML_CT)
         with patch("requests.get", return_value=mock_resp):
-            resp, _ = html.fetch(f"https://{target_apex}/")
+            page = html.fetch(f"https://{target_apex}/")
+
+        assert page.response is mock_resp
+
+
+class TestWebpageCsrfSignals:
```
coolhandle01: I think all these tests should have gone into a new file, this file is now 5x as long as it was..

tools/html.py
coolhandle01: Good catch — I was incomplete in my earlier reply. `mimetypes.types_map` IS a canonical dict (keyed by **file extension**, e.g. `mimetypes.types_map[".html"] -> "text/html"`, `[".json"] -> "application/json"`, `[".js"] -> "application/javascript"`). So there's a stdlib source of truth, just keyed awkwardly for our case (we want to lookup by name, not by extension).

Using it would look like:

```python
import mimetypes
_HTML_CT = {"Content-Type": mimetypes.types_map[".html"]}  # "text/html"
_JSON_CT = {"Content-Type": mimetypes.types_map[".json"]}  # "application/json"
```

Cleaner-feeling but the `[".html"]` indirection is awkward for the reader at the constant-definition site. I'd weakly lean keep the literal `"text/html"`/`"application/json"` for the two constants we have — they're stable IETF strings (RFC 2046 / 8259) that don't drift. The `mimetypes` indirection earns its keep when you're *guessing the type from a filename you already have* — exactly what `tools/cloud/services.py` could use later for the sensitive-files probe.

**On the "is incorrect MIME type a finding" side question — yes, real attack class.** Two flavours:

1. **MIME sniffing / content sniffing.** Old IE (and modern browsers in some contexts) ignore the server's `Content-Type` and try to detect the actual content. An attacker uploads `cute.png` containing HTML; the browser sniffs it as HTML and executes script in the origin's context → stored XSS. Mitigation: `X-Content-Type-Options: nosniff` header. Already covered by `webapp_headers.py` `check_security_headers` (nosniff is in the missing-header list).

2. **Mismatched type on user-uploaded files.** Serving user-controlled content with `Content-Type: text/html` when it should be `application/octet-stream` (downloads) or `text/plain` (raw text) → same stored-XSS story. This is what file-upload-XXE issues (#75 SVG, #76 OOXML) bleed into — server sending the wrong Content-Type back on download. Worth filing as a sibling issue under the file-upload umbrella; could probe with a polyglot upload + GET-the-thing-back + assert `Content-Type` matches the declared type. Defer to a follow-up since it's authenticated-upload territory (same session-management blocker as #75 / #76).

---

tests/tools/test_webpage.py
```diff
@@ -0,0 +1,324 @@
...
+    """<link rel="stylesheet" href="..."> parsing on the Webpage shape."""
+
+    def test_picks_up_stylesheets(self, make_page) -> None:
+        page = make_page(body='<link rel="stylesheet" href="/style.css">')
+        assert len(page.stylesheets) == 1
+        assert page.stylesheets[0].url == "/style.css"
+
+    def test_ignores_non_stylesheet_links(self, make_page) -> None:
+        page = make_page(
+            body=('<link rel="icon" href="/favicon.ico"><link rel="canonical" href="/canonical">'),
```
coolhandle01, line 391: Then again, page.favicon would give us something to feed into shodan one day. Could you add a cached property for the favicon too, please? 😇

tests/tools/test_webpage.py
```diff
@@ -0,0 +1,324 @@
...
+            # Unambiguous cookie signals
+            ({"csrftoken": "abc"}, "", {Framework.django}),
+            ({"_xsrf": "abc"}, "", {Framework.tornado}),
+            # Unambiguous meta signal
+            (None, '<meta name="csrf-param" content="auth_token">', {Framework.rails}),
+            # Ambiguous - typed channel stays empty even though page_protected flips
+            ({"XSRF-TOKEN": "abc"}, "", set()),
+            (None, '<meta name="csrf-token" content="abc">', set()),
+            # No signals
+            (None, "<html></html>", set()),
```
coolhandle01, line 118: I'm not seeing angularjs, angular, react/next.js/Vue, .. bootstrap, material.js etc. it feels like we've got some @pentest_tools to attack the js surface but nothing to identify the js Framework at the surface..

.. if OA and VR can identify vulnerable third party js (and CSS), PT can surely exploit it so long as there is a tool to do it.

I remember angularjs had `ng`-prefixed attributes and some ceremony around onload=. I'm sure they're all got they're own bootstrapping "fingerprints" that we could detect (aside from checking the CSS and js imports, especially is has_integrity is true. Just 'cos your bootstrap has SRI doesn't mean it's not some vulnerable version!

So, Framework could mean laravel, rails, but it could also mean React, SPA.

coolhandle01: Good push — landed in `df956eb`.

**Framework enum** now covers:

- **Server-side**: django, laravel, rails, spring, tornado
- **Client-side / SPA**: angular (modern 2+), **angularjs** (legacy 1.x — kept distinct per your "ng-prefixed attributes and ceremony around onload" observation; the two share almost no attack surface), nextjs, **react**, **vue**
- **CSS framework**: **bootstrap** — your "SRI on bootstrap.min.css doesn't make the version safe" point is the rationale. Per-version XSS / CVE bookkeeping is the same shape as a JS framework's, so it earns a member here rather than a separate `CssFramework` enum.

**Detection** is now three streams that all land in `Webpage.frameworks`:

1. **Response cookies** — `csrftoken` (Django), `_xsrf` (Tornado) — already there.
2. **HTML `<meta>` tags** — `csrf-param` (Rails) — already there.
3. **`<script>` + `<link>` URL substrings** — new. `_FRAMEWORK_URL_SIGNALS` in `tools/html.py` maps `@angular/` → Angular, `angular.js` → AngularJS (order matters; scoped-package check fires first so modern Angular's bundled scripts don't get classified as 1.x), `react.production`/`react-dom` → React, `/_next/` → Next.js, `bootstrap.` → Bootstrap, etc. Matches case-insensitively against both the raw href/src and the resolved absolute URL so CDN refs catch even when the relative form doesn't.

**Deliberately NOT in v1:**

- HTML-attribute heuristics for AngularJS (`ng-app`, `ng-controller`, `ng-model`). Could add via `_FRAMEWORK_HTML_ATTR_SIGNALS` later; URL-pattern detection covers most real-world cases.
- Version extraction. The URL signal flags "Bootstrap present" but not "Bootstrap 4.3.1". The natural place for version parsing is the next-iteration nuclei / CVE-lookup probe — the typed `Framework` channel hands it the *what*, the version probe figures out the *which*.
- Material UI / Tailwind / Svelte / Ember. Add when there's a probe that targets them; the catalogue stays append-only.

---

coolhandle01:
#187 is going to rug-pull this in a good way

Issue #183, opened 2026-05-30
feat: async-optimise @cyber_tool for I/O-bound wrappers (fan-out support)


coolhandle01:
## Motivation

Carved out of **#176**, whose async fan-out slice never landed (the nmap refactor stayed synchronous). Async is not nmap-specific: it is a cross-cutting capability `@cyber_tool` should grow once, so every I/O-bound recon wrapper can fan N concurrent calls instead of blocking the agent turn per host.

The groundwork is already visible in the typing: `SquadTool.func` is declared `@property` precisely because CrewAI's `Tool.func` is typed `Callable[P, R | Awaitable[R]]` (`squad/__init__.py:63`). The `Awaitable` half is the seam this issue uses - CrewAI's `Tool` model already accommodates async funcs at the type level; what is missing is the squad-side decorator support and a confirmed runtime pattern.

### 1. `@cyber_tool` recognises `async def`

Extend `cyber_tool` (`squad/__init__.py:77`) to accept an `async def` wrapped function and produce a working CrewAI tool. Spike CrewAI's actual stance first:

- If `crewai.tools.tool(name)` already drives an async `func` correctly (awaits it on the tool-call path), passthrough - no wrapping needed.
- If not, wrap with a sync shim (`asyncio.run(...)` / `anyio.from_thread`-style bridge) at the decorator boundary so the agent-facing surface stays synchronous while the wrapper body is `async def`.

Whatever lands here is the template every subsequent async tool follows. Document the chosen pattern in the `cybersquad-tool` skill as part of this PR.

### 2. First consumer: nmap fan-out

Convert `tools/recon/nmap/scanner.py:nmap_scan` to `async def` and fan one subprocess-async nmap call per host (or per host-bucket, depending on what nmap's own parallelism gives before we double up). This is the concrete win #176 specified; landing it here closes that loop.

### 3. Other I/O-bound candidates (convert where it pays)

Audit the recon/enrichment surface and convert the wrappers whose latency is dominated by independent per-target I/O:

- `tools/recon/rdap.py` - the `lookup_rdap_many`-style batch path (per-IP HTTP, per-RIR soft rate caps; honour `adaptive_sleep` cadence even when concurrent).
- httpx-based probe / tech-detect passes that already iterate per endpoint.
- Team Cymru / dnsx batch lookups where the binary or endpoint does not already parallelise internally.

Not every tool benefits - CPU-bound or already-batched binaries stay sync. "where appropriate" is load-bearing.

[...]

Issue #180, opened 2026-05-28
feat: RDAP lookup tool - structured registrant entry for IPs and domains


coolhandle01:
## Motivation

Team Cymru's bulk whois (landed in #176's third slice as `tools/recon/asn.py:lookup_asn`) gives us **`AsnRecord(ip, asn, prefix, country, organisation)`** — coarse but cheap. The `organisation` field is a single string ("GOOGLE, US"), useful for grep but not for structured downstream work.

**RDAP (RFC 7483)** is the modern replacement for whois. JSON over HTTP. The five RIRs (ARIN, RIPE, APNIC, LACNIC, AfriNIC) all serve RDAP at well-known endpoints, and `rdap.org` is a meta-router that redirects to the right RIR for a given query. Returns a structured registrant entry per the RFC schema: registrant org name + address, registrant abuse contact, registrar, registration date, last-changed date, status (active / suspended / etc.), nameservers (for domain queries).

That structured shape is the "professional registrant record" cybersquad wants persisting as amass `SourceProperty` / `SimpleProperty` values when amass lands.

### `tools/recon/rdap.py`

```python
def lookup_rdap(target: IPAddress | Hostname) -> RdapRecord | None: ...
```

* HTTP GET to `https://rdap.org/<query>` (router; redirects to the right RIR's RDAP endpoint).
* Pure HTTP via `tools.http` — no new binary, no new dep.
* Returns ``None`` on 404 / network failure; degrades gracefully like `lookup_asn`.

Optional batching helper: `lookup_rdap_many(targets)` calls in sequence with the operator's `adaptive_sleep` cadence. RDAP endpoints have per-IP rate limits; honour them.

### `models/network.py:RdapRecord`

```python
class RdapRecord(BaseModel):
    target: IPAddress | Hostname          # what was queried
    registrant_organisation: str | None   # the registered legal entity
    registrant_country: str | None        # ISO 3166-1 alpha-2
    registrar: str | None                 # which registrar / RIR registered it
    registered_at: datetime | None
    last_changed_at: datetime | None
    status: list[str]                     # ["active", "client transfer prohibited", ...]
    abuse_email: str | None               # RFC 7485 abuse-contact-discovery output
    nameservers: list[Hostname]           # for domain queries; empty for IP queries
```

[...]

PR #184, opened 2026-05-30
fix(recon): pass hostnames (not URLs) from run_recon to nmap

coolhandle01:
## What

`run_recon` built `live_hosts` from `Endpoint.url` — full `https://host/path` URLs — and handed them to `port_scan` → nmap. nmap scans hosts / IPs, not URLs, so it failed to resolve the URL targets: the Initial Sweep's port scan returned nothing useful and `open_ports` ended up keyed by URL instead of host, breaking the host-keyed lookups the `Recon Open Ports` slicer does downstream. (`tools/recon/__init__.py:153` already used the correct `urlparse(ep.url).hostname` idiom elsewhere — line 140 just didn't.)

## How

- Extract the hostname via the existing `host_of()` helper before `port_scan`, dropping hostless URLs and keeping the live-endpoint (`< 500`) filter.
- Tighten `nmap_scan` / `port_scan` signatures from `list[str]` to `list[FQDN]` to document that nmap takes concrete hosts — the type that would have surfaced this mismatch at the call site.

## Why this surfaced

Came out of an audit of the OSINT Analyst tool surface for loose `list[str]` host inputs (originally chasing whether wildcard / SAN entries like `*.example.com` could reach a scanner). The agent-facing `@cyber_tool` boundary turned out to be correctly typed already — every host field is `FQDN`-typed (which rejects wildcards), and wildcards are resolved upstream by subfinder, not by the type system. The only genuine defect was this wrong-*value* (URL vs host) bug in the legacy sweep path, not a wrong-*type* at the boundary.

## Tests

- New regression test `test_port_scan_receives_hostnames_not_urls` asserts `port_scan` receives the bare hostname (scheme + path stripped) and that 5xx endpoints are filtered out. Verified it fails on the pre-fix code (`['https://api.example.com/admin']`) and passes on the fix.
- `ruff`, `mypy`, and the recon / nmap / OSINT-analyst test suites all green.

## Scope

Single concern (nmap host-input correctness). Not in scope: the async fan-out work (tracked in #183) or the `#176` / `#180` closures.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JnQfpk4c3RrmXpLWkpjXVo)_

### DIFF
--- a/tools/recon/nmap/scanner.py
+++ b/tools/recon/nmap/scanner.py
@@ -119,7 +119,7 @@ def nmap_scan(
     return NmapScanResult(mode=mode, hosts=parsed, evidence_path=evidence_rel)
 
 
-def port_scan(hosts: list[FQDN]) -> dict[str, list[int]]:
+def port_scan(hosts: list[FQDN]) -> dict[FQDN, list[int]]:
     """Backwards-compatible thin shim over ``nmap_scan``.
 
     Calls ``nmap_scan(mode=QUICK_PORTS, persist_evidence=False)`` and
@@ -133,7 +133,7 @@ def port_scan(hosts: list[FQDN]) -> dict[str, list[int]]:
         scripts=NmapScripts.NONE,
         persist_evidence=False,
     )
-    out: dict[str, list[int]] = {host: [] for host in hosts}
+    out: dict[FQDN, list[int]] = {host: [] for host in hosts}
     for host_result in result.hosts:
         open_ports = [s.port for s in host_result.services if s.state == "open"]
         out[host_result.host] = open_ports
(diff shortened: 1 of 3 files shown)


tools/recon/nmap/scanner.py
```diff
@@ -118,7 +119,7 @@ def nmap_scan(
     return NmapScanResult(mode=mode, hosts=parsed, evidence_path=evidence_rel)
 
 
-def port_scan(hosts: list[str]) -> dict[str, list[int]]:
+def port_scan(hosts: list[FQDN]) -> dict[str, list[int]]:
```
coolhandle01:
```suggestion
def port_scan(hosts: list[FQDN]) -> dict[FQDN, list[int]]:
```

PR #185, opened 2026-05-30
refactor(#83): CPE-centred recon - delete Technology, annotate OAM Service assets, NVD client

### LINKED ISSUE #83: feat: tools/pentest/framework/ - framework-specific vulnerability probes

coolhandle01:
Closes #83.

## Why

#83 had drifted into a hand-maintained `Technology` / `TechnologyCategory` ontology with a ~90-row name→category catalogue that was *also* acting as an allow-list — silently dropping any tech string it hadn't catalogued. That re-derived data wappalyzer / NIST already own, and quietly violated "store the detecting tool's own string."

This rewrite re-orients recon's tech signal around the **OWASP Open Asset Model**: a detected technology is a *property on an asset*, not a bespoke model. We classify nothing, invent no names, and carry each tool's own identifier — plus, where a tool hands us one, the NIST **CPE** (the authoritative product key into NVD's CVE data).

## What changed, in slices (one concern per commit)

1. **Capture nmap CPEs** — nmap `-sV` emits a CPE per matched service (legacy 2.2 URI binding); we were dropping them. New `tools/cpe.py` (via the `cpe` library) normalises to the canonical CPE 2.3 formatted string; `NmapService.cpe` carries it.
2. **Drop the catalogue** — delete `_CATALOGUE`; `coerce_technologies` becomes a near-passthrough that keeps *every* observed string (full situational awareness), classifying nothing.
3. **Delete `Technology`** — the typed-tech model and its `TechnologyCategory` enum go entirely (no tool emits categories; nothing consumed them). Tech facts live as properties on the OAM asset instead. `Service` gains `cpe` + `detected_by` (OAM `SourceProperty`) and drops its redundant `technologies` list; `Endpoint` keeps its raw httpx `technologies: list[str]` (the tool's own vocabulary).
4. **Land the data in the rundir** — `Discover Host Services` (the nmap deep-scan tool) now returns OAM `Service` assets and writes them to `<run_dir>/hosts/<fqdn>/services.json` — the on-disk form of a future amass `Service`-node upsert. `NmapHostResult` no longer crosses the agent boundary (scanner-internal types stop leaking through `@cyber_tool`).
5. **Rename the OA tool surface** to a small `Discover` / `List` OAM-noun taxonomy (was a mix of Run / Recon / Probe / Detect / Deep Scan). `Discover` = active recon that produces assets; `List` = read/slice the collected graph. Display name + function + args_schema renamed together; `tools/recon/*` implementation names untouched.
6. **Consolidated NVD client** (`tools/nvd.py`) — one typed client for the CVE + CPE APIs: `cves_for_keyword`, `cves_for_cpe` (the nmap-CPE → CVEs path, pairs with `Service.cpe`), and `search_cpes` (the CPE dictionary). Cached in-process, api-keyed from the existing `config.scan.nvd_api_key`, degrades to empty on error. The pre-existing keyword-only `lookup_cve` is absorbed and deleted — one client, not two.

Plus the review round: the OA discovery/curation tools now return **typed OAM values** (`list[FQDN]`, `list[HttpUrl]`) and validate through Pydantic `TypeAdapter` (a bare `FQDN(x)` call does *not* run the validator — it's cosmetic; this makes the typed returns real).

## OA tool renames (display name)

| Before | After |
|---|---|
| Deep Scan Host | Discover Host Services |
| Probe FQDNs | Discover Webpages |
| Detect Takeover Candidates | Discover Takeover Candidates |
| LLM Endpoint Detection | Discover LLM Endpoints |
| Historical URL Discovery | Discover Historical URLs |
| Certificate Transparency Lookup | Discover Subdomains |
| Recon Subdomains / Endpoints / Open Ports | List Subdomains / Endpoints / Open Ports |
| Uncovered Hosts | List Uncovered Hosts |

Unchanged: Run Initial Sweep, Lookup IP Assets, Lookup RDAP for ASN, Lookup CWE, Lookup OWASP Guidance, Annotate Host, Finalise Recon.

## Notable

- **New dependency: `cpe` (LGPLv3)** — the repo's first copyleft dep. Used purely as an imported parser (no modification / static linking), so obligations are light; flagged in `pyproject.toml` for the record. Happy to hand-roll the 2.2→2.3 conversion instead if you'd rather keep the dep list permissive.
- **`search_cpes` / `cves_for_cpe` have no agent wiring yet** — low-level only. They're the substrate for the (separate, still-parked) wappalyzer-name → CPE resolution and the CVE-enrichment step.

## Tests

Full unit suite, `ruff` (+ format), `mypy`, `bandit`, and the per-PR `diff-cover` ratchet (100% of changed lines) all green. All external HTTP (NVD) mocked; no live calls.

## Follow-ups (not in this PR)

- wappalyzer name → CPE resolution for web tech (decision parked: whether structured per-tech CPE goes back onto `Endpoint`, or web tech stays string-keyed with CPE precision reserved for nmap `Service`s).
- `@cyber_tool` async fan-out — #183.


squad/osint_analyst/discovery.py
```diff
@@ -227,61 +227,61 @@ class _CertTransparencyArgs(BaseModel):
     )
 
 
-@cyber_tool("Certificate Transparency Lookup", args_schema=_CertTransparencyArgs)
-def cert_transparency_tool(domain: FQDN) -> list[str]:
+@cyber_tool("Discover Subdomains", args_schema=_DiscoverSubdomainsArgs)
+def discover_subdomains_tool(domain: FQDN) -> list[str]:
```
coolhandle01:
```suggestion
def discover_subdomains_tool(domain: FQDN) -> list[FQDN]:
```

Shouldn't this be returning a list of more FQDNs?

squad/osint_analyst/discovery.py
```diff
@@ -227,61 +227,61 @@ class _CertTransparencyArgs(BaseModel):
...
-            " Probe FQDNs to confirm liveness before annotating."
+            " Discover Webpages to confirm liveness before annotating."
         ),
     )
 
 
-@cyber_tool("Historical URL Discovery", args_schema=_HistoricalUrlsArgs)
-def historical_urls_tool(domain: FQDN) -> list[str]:
+@cyber_tool("Discover Historical URLs", args_schema=_DiscoverHistoricalUrlsArgs)
+def discover_historical_urls_tool(domain: FQDN) -> list[str]:
```
coolhandle01:
```suggestion
def discover_historical_urls_tool(domain: FQDN) -> list[Endpoint]:
```

Shouldn't this be returning a list of URLs (I guess Endpoints for us) ?

tests/tools/pentest/test_vuln_tools.py
```diff
@@ -263,46 +266,35 @@ def _nvd_response(self):
...
-        with patch("requests.get", return_value=mock_resp):
+        with patch("tools.nvd.http.get", return_value=mock_resp):
             results = lookup_cve("log4shell")
 
-        assert len(results) == 1
-        assert results[0]["id"] == "CVE-2021-44228"
-        assert results[0]["cvss_score"] == 10.0
-        assert results[0]["description"] == "Log4Shell RCE"
+        # The shim still hands back plain dicts (not typed CveEntry) for
+        # back-compat with existing callers.
```
coolhandle01: Let's not leave technical debt, there's no real benefit to this half-implementation, it's just mess to clean up later

tests/tools/recon/httpx/test_probe_endpoints.py
coolhandle01: should be using fixtures for urls

tests/test_squad_osint_analyst.py
```diff
@@ -267,35 +269,35 @@ def test_lookup_owasp_tool(self) -> None:
...
+    def test_discover_historical_urls_tool(self, target_apex) -> None:
+        from squad.osint_analyst import discover_historical_urls_tool
 
         sentinel = [f"https://{target_apex}/old"]
         with patch(
             "squad.osint_analyst.discovery.historical_urls",
             return_value=sentinel,
         ) as m:
-            result = historical_urls_tool.func("example.com")
+            result = discover_historical_urls_tool.func("example.com")
```
coolhandle01: fixture please

tools/pentest/triage.py
```diff
@@ -29,46 +24,10 @@ def is_in_scope(finding: RawFinding, programme: Programme) -> bool:
 def lookup_cve(keyword: str) -> list[dict]:
     """Query the NVD API for CVEs matching a keyword to validate CVSS scores.
 
-    Returns up to 5 results. Returns [] on any error so the pipeline is never
-    blocked.
+    Back-compat shim over ``tools.nvd.cves_for_keyword`` - the single NVD
+    client now lives in ``tools/nvd.py``. Kept returning the historical
+    ``list[dict]`` shape so existing callers (the VR's ``NVD CVE Lookup``
+    wrapper) and their tests are undisturbed. Up to 5 results; ``[]`` on any
```
coolhandle01: bit sloppy really tbh, back-compat for who? for nobody :) plz complete the implementation

squad/osint_analyst/curation.py
```diff
@@ -199,8 +199,8 @@ class _UncoveredHostsArgs(BaseModel):
     )
 
 
-@cyber_tool("Uncovered Hosts", args_schema=_UncoveredHostsArgs)
-def uncovered_hosts_tool(attack_graph_path: str = "attack_graph.json") -> list[str]:
+@cyber_tool("List Uncovered Hosts", args_schema=_ListUncoveredHostsArgs)
+def list_uncovered_hosts_tool(attack_graph_path: str = "attack_graph.json") -> list[str]:
```
coolhandle01: should probably be returning `list[FQDN]` ?

PR #192, opened 2026-06-03
feat(tasks): guardrail the select -> recon handoff (#168)

### LINKED ISSUE #168: feat(squad): Task Guardrails on `select → recon` (function variant) + TaskOutput fixture
Carve-off from #144 — concrete grounding from the [CrewAI Task Guardrails quickstart](https://github.com/crewAIInc/crewAI-quickstarts/blob/main/Guardrails/task_guardrails.ipynb).

## The shape

```python
def validate_blog_content(result: TaskOutput) -> Tuple[bool, Any]:
    if word_count > 50:
        return (False, "Blog content exceeds 50 words")
    return (True, result.raw.strip())

Task(..., guardrail=validate_blog_content, max_retries=4)
```

`TaskOutput → (bool, value_or_error)`. On `False` the error string is fed back to the agent and the task re-runs up to `max_retries`. **Function guardrail — no LLM judge, no extra LLM spend.** This is the path we want; the LLM-judge variant earns its keep on free-text outputs we don't have.

## First target: `select → recon`

A malformed `select.json` derails everything downstream and is detected late today. Concrete sketch:

```python
def validate_select_output(result: TaskOutput) -> Tuple[bool, str]:
    try:
        select = SelectResult.model_validate_json(result.raw)
    except ValidationError as e:
        return (False, f"select.json failed schema validation: {e}")
    if not select.programme_handle:
        return (False, "select.json must include programme_handle")
    return (True, result.raw)
```

~10 lines, no new infra — we already have the typed `SelectResult` model so the validation is "construct the Pydantic model"; the retry budget makes failure actionable (the agent gets told *why* it failed and tries again).

## TaskOutput BDD fixture — required, not nice-to-have

With the guardrail signature being `TaskOutput → (bool, Any)`, a `TaskOutput` fixture is the **unit-test surface** for guardrail functions, not just BDD scaffolding:

```python
def test_select_guardrail_rejects_missing_programme_handle(make_task_output):
    bad = make_task_output(raw='{"selected_at": "2026-05-26"}')
    ok, msg = validate_select_output(bad)
    assert not ok
    assert "programme_handle" in msg
```

Same parallel as the `Programme` / `dvwa_programme` fixtures in `tests/fixtures/programme.py` — give the test the real Pydantic type, not `MagicMock` with whatever attrs the test author thought to set. Belongs in `tests/fixtures/task_output.py`.

## Scope

- New module `squad/guardrails.py` (or inline in `tasks.py` for now) with `validate_select_output`.
- Wire into the `select` task in `tasks.py` via `Task(..., guardrail=validate_select_output, max_retries=2)`. Modest retry budget — guardrail failures should converge or fail loudly, not loop expensively.
- New fixture: `tests/fixtures/task_output.py` with `make_task_output` factory.
- Unit tests for `validate_select_output` against the fixture.
- BDD scenario covering the guardrail-rejection path (a malformed `select` triggers retry + recovery).

## Out of scope

- Other handoff boundaries (`recon → research`, etc.) — evaluate after `select → recon` lands and we know whether the pattern earned its keep.
- LLM-judge guardrails — the function variant is sufficient for our typed workspace artefacts.
- Event-listener observability for guardrail firings — sibling work; see the EventListener scaffolding thread in #144 comment 2. The two compose: function guardrails *enforce*, the event bus *surfaces* "guardrail rejected, retry 2/3" to the TUI.

## Skill updates

- `cybersquad-task` grows a "guardrails" section pointing at `validate_select_output` as the canonical pattern (function, not LLM judge; uses the existing typed model).
- `cybersquad-test-fixtures` grows a `TaskOutput` fixture entry next to the `Programme` one.

coolhandle01:
Closes #168.

## What

Adds a CrewAI *function* guardrail on the `select` task so a malformed or
missing programme selection is caught and fed back to the Programme Manager
for a bounded retry, rather than silently derailing `recon` downstream.

## How

- **`squad/guardrails.py` (new)** - `validate_select_output(result)`.
  Validates the *persisted workspace artefact*, not `result.raw`: it calls the
  same `current_programme()` that `recon` will call, so it checks the exact
  `<run_dir>/programme.json` the next agent reads. Distinguishes the failure
  modes with actionable messages - no programme written (RuntimeError /
  FileNotFoundError), schema-invalid JSON (ValidationError), and empty handle.
  On success it passes `result.raw` through unchanged so the `context=` chain
  is byte-for-byte identical to today.
- **`squad/__init__.py`** - `build_task` gains keyword-only `guardrail` and
  `max_retries` passthrough, both defaulting to `None` (CrewAI's own defaults),
  so every un-guarded task is unchanged.
- **`tasks.py`** - wires `select` with `guardrail=validate_select_output,
  max_retries=2`. No other handoff is guarded - the pattern earns its keep at
  `select -> recon` first.

## Tests

- **`tests/fixtures/task_output.py` (new)** - `make_task_output` factory,
  registered in `tests/conftest.py`. Hands the guardrail a real
  `crewai.TaskOutput` rather than a MagicMock.
- **`tests/squad/test_guardrails.py` (new)** - pass, missing programme.json,
  unbound run_dir, malformed JSON, empty handle, and raw passthrough.
- **`tests/test_tasks.py`** - asserts only `select` is guarded and that it
  carries `validate_select_output` / `max_retries == 2`.

## Skills

- `cybersquad-task` gains a guardrails section (function variant, validate the
  artefact not `result.raw`, wire via `build_task`, keep `max_retries` modest).
- `cybersquad-test-fixtures` documents `make_task_output`.

## Scope note

Unit-only by design. The BDD scenario that would share the `programme.json`
fixture is deliberately deferred rather than bootstrapping the repo's first
real-LLM BDD harness in this PR - the runtime guardrail stands on its own.

## CI parity (local, .venv @ 3.12.3)

ruff / ruff format / mypy (clean) / pylint (9.91, floor 9.88) / bandit clean /
pytest -m unit green / diff-cover 100% on changed production lines.

### DIFF
--- a/.claude/skills/cybersquad-task/SKILL.md
+++ b/.claude/skills/cybersquad-task/SKILL.md
@@ -69,6 +69,27 @@ triage = build_task(
 
 Always pass `human_input=hi` where `hi = config.human_input` (set via the `CYBERSQUAD_HUMAN_INPUT` env var). Never hardcode `True` or `False` - the toggle exists so production runs can be unattended while interactive runs gate at each step.
 
+## Guardrails: validate a handoff before it derails downstream
+
+A CrewAI *function* guardrail is `(TaskOutput) -> tuple[bool, Any]`: return `(True, value)` to pass `value` to the next task, or `(False, error)` to feed `error` back to the agent, which re-runs the task up to `max_retries`. Use the **function** variant (plain Python against our typed artefacts), not the LLM-judge variant - our workspace outputs are typed models we can just construct, so there is no free-text judgement to delegate and no extra LLM spend.
+
+Canonical example: `squad/guardrails.py:validate_select_output`, guarding `select -> recon`.
+
+**Validate the workspace artefact, not `result.raw`.** Inter-agent handoff flows through workspace files (see the section above), so the thing that actually derails recon is a malformed `<run_dir>/programme.json`, not the agent's prose answer. `validate_select_output` therefore calls the same `current_programme()` recon will call - validating the exact artefact the next agent reads - and only passes `result.raw` *through* (as the `True` value) so the `context=` chain is unchanged. Keying a guardrail off `result.raw` instead would validate the agent's free-text, which is fragile (agents wrap JSON in prose / fences) and checks the wrong surface.
+
+Wire it through `build_task`, never a bare `Task(guardrail=...)`:
+
+```python
+select = build_task(
+    "select", PROGRAMME_MANAGER, agents["programme_manager"],
+    human_input=hi,
+    guardrail=validate_select_output,
+    max_retries=2,
+)
+```
+
+`build_task`'s `guardrail` / `max_retries` params default to `None` (CrewAI's own defaults), so un-guarded tasks are unchanged. Keep `max_retries` modest - guardrail failures should converge or fail loudly, not loop expensively. Unit-test the guardrail directly with the `make_task_output` fixture (`tests/fixtures/task_output.py`) plus the rundir-staging fixtures (`run_dir` / `programme_in_workspace`); see `tests/squad/test_guardrails.py`. Other handoff boundaries (`recon -> research`, ...) stay un-guarded until the pattern earns its keep there.
+
 ## Upstream alignment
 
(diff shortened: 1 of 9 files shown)


coolhandle01:
Validating the model is good, we should validate that output.raw has some resemblance to the expected output too though.

